PR #15, opened 2024-11-16
Update the bitmask to 256 bits

samufi:
256 are required if we want to support 256 components.
The new implementation is much cleaner as it builds on the SIMD functionality.


samufi:
Note: the test fails because world.mojo is in an unfinished state and contains syntactically wrong code. This will be fixed in an upcoming PR.

src/bitmask.mojo
```diff
@@ -1,36 +1,53 @@
...
+    fn __init__(inout self, bytes: SIMD[DType.uint8, Self.total_bytes]):
+        """Initializes the mask with the given bytes."""
+        self._bytes = bytes
 
+    @always_inline
+    fn __init__(inout self, bits: VariadicList[BitMask.IndexType]):
+        """Initializes the mask with the bits at the given indices set to True.
+        """
+        self._bytes = SIMD[DType.uint8, Self.total_bytes]()
+        for bit in bits:
```
mlange-42, line 26: Is `bits` like a list of bools? Asking to clarify whether the loop is really required...

mlange-42: Ah, I see, it is a list of indices...

src/bitmask.mojo
```diff
@@ -40,96 +57,81 @@ struct BitMask(Stringable):
...
-        self.hi = 0
+        self._bytes = 0
 
+    @always_inline
     fn contains(self, other: Self) -> Bool:
         """Reports if the other mask is a subset of this mask."""
-        return ((self.lo & other.lo) == other.lo) and (
-            (self.hi & other.hi) == other.hi
-        )
+        return ((self._bytes & other._bytes) == other._bytes).reduce_and()
```
mlange-42, line 109: You may want to benchmark this (and other uses of `reduce...`) against a version not using loops/reduce. In my experience, loop unrolling can make a big difference, and I doubt that the Mojo compiler is already mature enough to do this optimization.

samufi: Well, there is an explicit syntax for loop unrolling already. However, since I am using SIMD here, this is not even implemented in Mojo but in MLIR (the intermediate pre-compiler language comparable to LLVM). Hence this exploits hardware vectorization and all that stuff. I hardly believe that my custom code could beat this.
I'd be happy to see this benchmarked, but I have not figured out a good way to do this.

src/bitmask_test.mojo
```diff
@@ -108,6 +127,17 @@ fn test_bit_mask_without_exclusive() raises:
     assert_false(excl.matches(BitMask(UInt8(1), UInt8(2), UInt8(3), UInt8(13))))
 
 
+fn test_bi_mask_eq() raises:
```
mlange-42, line 131:
```suggestion
fn test_bit_mask_eq() raises:
```

src/bitmask.mojo
mlange-42: Just write an interim alternative method that does it like in Arche, just for benchmarking to see if it makes a difference? Well, that is probably not what you mean.

mlange-42 (approved):

src/bitmask_test.mojo
samufi: Fixed.

PR #17, opened 2024-11-25
Implement a BitMask graph for quick BitMask lookup

samufi:
The graph facilitates quick access of items identified by bitmasks if an item with a somewhat similar bitmask is known.

The graph will be used for quick Archetype access. It is implemented via a list of inline arrays and nodes are identified via their index in this list.

During the implementation of this graph, a bug in the hash function of SIMD was detected, which is why we circumvented this via a naively implemented dict, which will be replaced by the Mojo dict in due time.


src/graph.mojo
```diff
@@ -0,0 +1,163 @@
...
+    # The index indicating a non-established link.
+    alias null_index = -1
+
+    # The value stored in the node.
+    var value: DataType
+
+    # The indices of the neighbouring nodes.
+    # The node at index i difffers from the
+    # current node by the i-th bit.
+    var neighbours: InlineArray[Int, 256]
```
mlange-42, line 27: Ah, I think now I understand what you mean by the flip simplification... holding only one list of neighbors instead of one for addition and one for removal. Just logic that this is possible, ut didn't think that far.

mlange-42: Used the opportunity to change the implementation in Arche accordingly.

src/graph.mojo
```diff
@@ -0,0 +1,163 @@
...
+        Args:
+            from_node_index: The index of the node from which the link is created.
+            changed_bit:     The index of the bit that differs between the nodes.
+
+        Returns:
+            The index of the node to which the link is created.
+        """
+        new_mask = self._nodes[from_node_index].bit_mask
+        new_mask.flip(changed_bit)
+        optional_to_index = self._map.get(new_mask)
```
mlange-42, line 114: I guess you plan to call this method only if there is not a link yet (which is what the method name implies)? Otherwise, you should check that here. In Arche, there is no separate type for the graph, and I do this all in one method, where it looks like this:

```go
if next, ok := curr.neighbors.Get(id.id); ok {
    curr = next
} else {
    next, _ := w.findOrCreateArchetypeSlow(mask, relation, hasRelation)
    next.neighbors.Set(id.id, curr)
    curr.neighbors.Set(id.id, next)
    curr = next
}
```

src/graph.mojo
```diff
@@ -0,0 +1,163 @@
...
+
+        Returns:
+            The index of the node differing from the start node by the given indices.
+        """
+        var current_node = start_node_index
+        for mask_index in different_bits:
+            var next_node = self._nodes[current_node].neighbours[
+                int(mask_index)
+            ]
+            if next_node == Self.null_index:
```
mlange-42, line 149: Ah, this resolves my previous comment.

It would probably be good to document this, here and in `create_link`.

samufi: Is it *necessary* to do this check? I think in the worst case, `create_link` would be less efficient than possible, but it will not be wrong even if it is called on an existing link.

src/graph.mojo
```diff
@@ -0,0 +1,163 @@
...
+        """Creates a link between two nodes.
+
+        Args:
+            from_node_index: The index of the node from which the link is created.
+            changed_bit:     The index of the bit that differs between the nodes.
+
+        Returns:
+            The index of the node to which the link is created.
+        """
+        new_mask = self._nodes[from_node_index].bit_mask
```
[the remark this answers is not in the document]
samufi: Yes, `=` typically invokes the copy constructor unless the thing's origin is transferred, which would invoke a `move`. `BitMask`, however, is a trivial register-passable type, meaning it is never passed by reference but always by copy. Hence, there is no move init.

src/graph.mojo
mlange-42: You are right. When commenting above, I had not yet seen the check here. So all fine, I would just document here "creates a new node and links it if not already present", and "does not check whether the link is already established" in `create_link`.

samufi: Well, but even if I did *not* check it, would it be wrong? 
I am happy to include the comments, though!

mlange-42 (approved):

PR #20, opened 2024-12-10
Fix entity creation; minor fixes

samufi:
Fix the process of creating entities and adding them to archetypes.

Fix minor issues raised in the review process

### DIFF
--- a/src/world.mojo
+++ b/src/world.mojo
@@ -116,6 +116,7 @@ struct World:
 
         return archetype_index
 
+    @always_inline
     fn new_entity(inout self) raises -> Entity as entity:
         """Returns a new or recycled [Entity].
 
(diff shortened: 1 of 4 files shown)


src/world.mojo
```diff
@@ -222,7 +223,7 @@ struct World:
     @always_inline
     fn _check_locked(self) raises:
         """
-        Checks if the world is locked, and panics if so.
+        Checks if the world is locked, and raises if so.
```
mlange-42, line 226: Mid-term, it would probably be good to create a specific error type here.

src/world.mojo
```diff
@@ -507,7 +508,7 @@ struct World:
     @always_inline
     fn is_locked(self) -> Bool:
         """
-        Rreturns whether the world is locked by any queries.
+        Returns whether the world is locked by any queries.
```
mlange-42, line 511: I would probably raise a `NotImplementedError` here. But you probably already use it somewhere... You should definitely implement this before starting queries.

mlange-42 (approved):
Comments are just suggestions for future MRs.

mlange-42:
Looks like the removed `archetype.has_relation()` is still in the tests.

src/world.mojo
samufi: Agreed. Short-term mojo does not have specific errors (because they don't have inheritance)

src/world.mojo
samufi: I will implement the lock prior to queries. If I raise an error here, though, testing will not be possible.

samufi:
> Looks like the removed `archetype.has_relation()` is still in the tests.

Fixed.

PR #21, opened 2024-12-18
Comptime lookup

samufi:
Make the lookup of component IDs compile time.

This makes the code faster and less error-prone.

### DIFF
--- a/src/component.mojo
+++ b/src/component.mojo
@@ -6,6 +6,7 @@ from stupid_dict import SimdDict, StupidDict as Dict
 from types import get_max_uint_size, TrivialIntable
 from memory import UnsafePointer
 from bitmask import BitMask
+from sys.intrinsics import _type_is_eq
 
 
 trait IdentifiableType:
@@ -17,7 +18,7 @@ trait IdentifiableType:
         ...
 
 
-trait ComponentType(IdentifiableType, Movable):
+trait ComponentType(Movable):
     pass
 
 
@@ -107,7 +108,7 @@ struct ComponentReference[is_mutable: Bool, //, origin: Origin[is_mutable]]:
         return self._id
 
 
-struct ComponentManager:
+struct ComponentManager[*component_types: AnyType]:
     """ComponentManager is a manager for ECS components.
 
     It is used to assign IDs to types and to create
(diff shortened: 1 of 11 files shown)


src/world_benchmark.mojo
```diff
@@ -48,7 +38,7 @@ struct Velocity(ComponentType):
 
 
 fn benchmark_new_entities_10_000() raises:
-    world = World()
+    world = World[Position, Velocity]()
```
mlange-42, line 41: Can't do an in-depth review before next week, but is this really required with the new implementation?

IMO the requirement to specify all component types at world creation really hurts modularity and decoupling.

mlange-42: In any case where you have different model versions that require different components. With your implementation, you will always need to know all possible types in the place where you create the world, potentially adding unused types. Without this requirement, you could just use them where needed.

As an example, see my BEEHAVE implementation [beecs](https://github.com/mlange-42/beecs). It is basically a library, and people who want to extend or alter it just import it and do their business. With the requirement discussed, this would not be possible as soon as new components are required. You would have to change the library code itself, preventing the import and requiring a code copy. Consequently, updating for fixes or improvements to the library would be more tedious then necessary, or you may just miss them entirely.

Alternatively, the library could be written in a way that users have to create and inject the world. But that would burden users with the requirement for the respective knowledge.

ATM I can't come up with a use case that would be made completely impossible, but certain things would definitely be less convenient and ergonomic than possible.

mlange-42: This looks like it would solve the problem. Not sure about recompilation in Mojo, but in Go this would be the case too, as static linking is the default.

Providing both options also sounds good. Actually, I think performance of component registration is not an issue anyway, as it happens only once in each run.

What do you mean by inheritance in this context? The philosophy of ECS is "composition over inheritance", so it would generally not be used for anything related to entities or components. They won't contain any model logic anyway, as this is implemented in systems.

samufi: For inheritance, I was thinking about extended models. Say, you have a forest model and want to add more functionality to it. You may want to keep all the processes but add new processes and components. Then you could inherit from the original forest model and add further processes without duplicating code. (I am thinking in an OOP framework, but of course one could also do that differently)

mlange-42: Yeah, you should definitely do this differently with ECS. Actually, the "original forest model" is just a list of systems (i.e. sub-models). To extend or modify, one would amend or change this list.

Taking `beecs` as an example again, [this](https://github.com/mlange-42/beecs/blob/0eee3be6e419c320800016a3fe67c438bbf31ad1/model/default.go#L15) is the original model. So there is not even a class or type that implements the actual model. It is all just generic, modular and plugged together.

IMO you definitely reached a point where you should **use** an ECS in a little project, to know what you are developing for.

samufi: Agreed. I will do that as soon as I have the minimal functionality of the ECS.

src/component.mojo
```diff
@@ -116,84 +127,55 @@ struct ComponentManager:
...
 
-        Raises:
-            Error: If the component was not registered and the maximum number of components has been reached.
-        """
-        if T.get_type_identifier() in self._components:
-            return self._components[T.get_type_identifier()].id
-        else:
-            return self._register[T, False]().id
+        # This is unreachable.
+        return -1
```
mlange-42, line 164: Maybe raise an error here?

samufi: Then I would need to make the function raising. For an unreachable error, this would be a disproportionate cost.

mlange-42: Would you strictly need to do that? I just mean, because this is not an exception that is intended to be caught. So I would panic, but there is no such thing in Mojo.

src/component.mojo
```diff
@@ -116,84 +127,55 @@ struct ComponentManager:
...
+            len(VariadicList(component_types)) <= Self.max_size,
+            "At most " + str(Self.max_size) + " component types are allowed.",
+        ]()
 
-    fn _register[
-        T: ComponentType, check_existent: Bool = True
-    ](inout self) raises -> ComponentInfo as component_info:
-        """Register a new component type.
+    @always_inline
+    fn get_id[T: ComponentType](self) -> Self.Id:
```
mlange-42, line 143: Do I understand correctly that the `@parameter` stuff causes this to be compiled to multiple functions like this?

```mojo
fn get_id[Position](self) -> Self.Id:
    return 0

fn get_id[Velocity](self) -> Self.Id:
    return 1
```

src/component.mojo
```diff
@@ -116,84 +127,55 @@ struct ComponentManager:
...
         If the component does not yet have an ID, register the component.
-
-        Raises:
-            Error: If the component was not registered and the maximum number of components has been reached.
         """
-        try:
-            return self._components[T.get_type_identifier()]
-        except Error:
-            return self._register[T, False]()
+        return ComponentInfo.new[T](self.get_id[T]())
```
mlange-42, line 172: Not sure where this is used, but instantiation on each call might degrade performance. As you know all used component types, you could pre-generate these.

src/component.mojo
samufi: Yes. And the result should be inlined. That is, the function call should be replaced with 0 and 1, respectively.

src/component.mojo
samufi: This is used to create an array of type IDs and type sizes. These tuples are trivial. Therefore, I believe constructing them in place is not slower than reading them from an array.

src/component.mojo
mlange-42: Yeah, and it would not provide a stack trace. So `raise` without `raises` is not possible?

src/component.mojo
mlange-42: I was more after the `sizeOf` that is called in the constructor, rather than the tuple contruction.

src/component.mojo
samufi: No, raise requires raises. This should be literally unreachable, as the compile time statement _contains_type does exactly the same thing as the for loop. I would have put a false constraint, but the compiler is not smart enough to ignore constraints after a return.

src/component.mojo
samufi: `sizeOf ` is evaluated at compile time.

mlange-42 (approved):

PR #22, opened 2024-12-19
Create new / unified benchmarks

samufi:
Use the builtin benchmarking system. The limitation is that the output is fixed to ms. Therefore, we do not benchmark individual actions but a whole bunch. Clear documentation should facilitate interpretation of the results anyway. Using the default framework allows better presentation of the results and may in the future also overcome the current limitations.


mlange-42 (commented):
Does not compile. LGTM otherwise.

PR #23, opened 2024-12-19
Implement component manipulation

samufi:
Components can be added to entities and removed from them (or both in one go).

To that end, a major issue needed to be fixed, namely that new archetypes need to be created based on the components they have and not based on the components in which they differ from another archetype. 

Furthermore, update the mojo version, make performance optimizations, and improve some assertions and other minor things.


src/archetype.mojo
```diff
@@ -114,32 +157,37 @@ struct Archetype(CollectionElement, CollectionElementNew):
...
-            id = int(other._ids[i])
-            size = other._capacity * int(other._item_sizes[id])
-            self._data[id] = UnsafePointer[UInt8].alloc(size)
-            memcpy(
-                self._data[id],
-                other._data[id],
-                size,
-            )
+        @parameter
+        for i in range(BitMask.total_bits):
```
mlange-42, line 179: This relies on compile-time lookup and iteration I guess? Otherwise, this is going to be quite slow, so I would be cargful here. Do you really need this constructor?

src/archetype.mojo
```diff
@@ -150,30 +198,59 @@ struct Archetype(CollectionElement, CollectionElementNew):
...
+        """Returns whether the archetype contains entities."""
+        return bool(self._size)
+
     @always_inline
     fn get_node_index(self) -> UInt:
         """Returns the index of the archetype's node in the archetype graph."""
         return self._node_index
 
+    @always_inline
+    fn get_mask(self) -> BitMask:
```
mlange-42, line 250: This creates a copy (I guess), which you do not want in most places where this is going to be used. So I would change it to return a pointer/ref, or you should be very careful to use `_mask` instead in most of the `World` code.

I see that returning a copy seems more safe, but it should be used only internally, anyway.

src/archetype.mojo
```diff
@@ -114,32 +157,37 @@ struct Archetype(CollectionElement, CollectionElementNew):
...
-        self._item_sizes = other._item_sizes
-        self._entities = other._entities
-        self._ids = other._ids
-        self._node_index = other._node_index
+    fn __init__(
+        out self,
+        node_index: UInt,
+        mask: BitMask,
+        component_manager: ComponentManager,
+        capacity: UInt = 10,
```
mlange-42, line 165: As `reserve` doubles the capacity, I would use a power-of-2 value here.

src/archetype.mojo
```diff
@@ -242,19 +329,84 @@ struct Archetype(CollectionElement, CollectionElementNew):
...
+            self.assert_has_component(id)
         return self._get_component_ptr(index, id)
 
     @always_inline
     fn has_component(self, id: Self.Id) -> Bool:
         """Returns whether the archetype contains the given component id."""
         return bool(self._data[int(id)])
 
+    @always_inline
+    fn has_all_components[T: Intable](self, ids: InlineArray[T]) -> Bool:
```
mlange-42, line 379: I don't have a sufficient overview to see where you use this and the following method, but wouldn't the use of masks be more appropriate here?

src/archetype.mojo
```diff
@@ -222,13 +299,23 @@ struct Archetype(CollectionElement, CollectionElementNew):
         )
 
     @always_inline
-    fn set(inout self, index: UInt, value: ComponentReference) raises:
+    fn set[
+        assert_has_component: Bool = True
```
mlange-42, line 303: I don't have a sufficient overview to see where you use this, but currently I don't see why you need the check for a method that is used internally only. So I think this should not be called on an inappropriate archetype in the first place.

src/archetype.mojo
samufi: Bit mask is small enough to fit into a single SIMD register. I would therefore think that making a copy comes at almost zero cost. And indeed I just confirmed this is a benchmark (though I am not sure I can measure things below 0.5ns very accurately). Note that the @always_inline may help that no copy is made.

src/archetype.mojo
samufi: To me, internally means within a struct. Therefore, this is not used internally. If the check should not be done, we can use the parameter to disable it. 

This method is used to set components of an entity. I think it should be checked somewhere whether the archetype has these components. Don't you?

src/archetype.mojo
samufi: It is used to see if all the components to be removed exist (in `world._remove_and_add`. Using a mask would require us to create and fill a ask first. I agree that a mask would be faster otherwise.

src/archetype.mojo
mlange-42: Have something similar in Arche too. Just did a test, and for a single component, I can achieve a speedup from 200ns to 50ns by considering only as many bits as there are registered components. 

https://github.com/mlange-42/arche/pull/453

samufi: I have implemented this as well now. No difference in the benchmarks (was at 40ns already anyway).

src/archetype.mojo
samufi: Done.

src/archetype.mojo
mlange-42: By internal I mean that it is used only from within the library, and not intended for end users. They should not even be able to access the archetypes at all (without using underscore stuff at their own risk, of course).

When used by the world, which is the only legal use, how could the wrong archetype be selected?

IMO the same applies for the all- and any-components check in the comment above.

samufi: For example, if we remove a component that an entity does not have.
```python
entity = wold.new_entity(position)
world.remove[Velocity](entity)
```

src/archetype.mojo
mlange-42: For the same reason as detailed in the comment below, I think these methods should not even be required. Only the world is responsible for selecting archetypes, and should not rely on such an expensive check for that sake. End users should not be able to remove components from an inappropriate archtype.

src/archetype.mojo
samufi: The question is only where the check is done: in the world or in the archetype. The latter has less redundant code.

mlange-42: Ok, you currently use this in the `World`.

In Arche, I use mask manipulation instead. See [World.getExchangeMask](https://github.com/mlange-42/arche/blob/16e3c44903401c8fc8a5cb1c65207b5e919b49dd/ecs/world_internal.go#L533).
This also covers duplicate component addition or removal, which would require an additional check in your implementation.

mlange-42: > The question is only where the check is done: in the world or in the archetype. The latter has less redundant code.

See above... the question is how to do the check, not where.

src/archetype.mojo
samufi: The world selects the entity's archetype via the archetype pointer / index stored in the EntityIndex. It therefore does not check the existence of components. If then a component is to be removed from the entity that it does not have, the program should raise rather than having undefined behaviour. Of course, this is not needed in queries or whatever where the archetype is searched by a filter.

src/archetype.mojo
samufi: How would I do the check alternatively? I think I don't understand.

samufi: The (only) advantage of world.set is that it can work with multiple components. You can do things like
`world.set(entity, pos, vel, accell)`. I think this could be useful.

samufi: Otherwise in mojo
```
world.get[Comp](e) = comp
```
or 
```
world.get_ptr[Comp](e)[] = comp
```
or 
```
comp_ptr = world.get[Comp](e)
comp_ptr[] = comp
```

src/archetype.mojo
mlange-42: The different archetype constructors are a bit confusing, but finally you need the archetype's mask to construct it. Creating this mask by altering the mask of the old/original archetype by setting the added/removed bits should be faster than from scratch, and can efficiently cover these checks. And you get the checks for duplicate entries in the add/remove lists for free, which would be relatively costly otherwise. I think you don't have these checks yet.

But go ahead, I don't want to insist further here.

mlange-42 (approved):

src/archetype.mojo
samufi: I will investigate the case of duplicate components. If not covered already, I will add compile-time checks. 

I am not sure if the thing with altering the original archetype's mask works here, because I do not distinguish between added and removed bits but simply say which bits change. Therefore, I need to assert that the bits were set / unset before. How would you do it otherwise? First do the remove operations and do a bitmask contains with the original mask and then do the addition and do a reverse bitmask contains? That could indeed work. I'll have a look.

mlange-42: > I do not distinguish between added and removed bits but simply say which bits change

But isn't that just true for the archetype graph? In `_add_remove`, you know what is added and what is removed.

Mask manipulation and checks are in [World.getExchangeMask](https://github.com/mlange-42/arche/blob/b2a90dd817fa43dc200d5ba5ea99db8a5e48ebb1/ecs/world_internal.go#L533)

The check for duplicate entries is implicit, and therefore for free.

samufi: OK, I have inserted a compile time check for uniqueness of components. This is free and also checks the program at compile time, which is nicer.

Furthermore, I did the check for adding / removing existent / nonexistent components based on the mask, and I let the getters return (immutable) references to the mask instead of copies (though I still am not sure whether this makes any difference)

PR #24, opened 2025-01-01
Implement query

samufi:
Implement queries over entities.


samufi:
This PR is replaced by #28, which contains the same content but resolves merge conflicts with main.

PR #28, opened 2025-01-05
Implement query

samufi:
Implement queries over entities.


src/query.mojo
```diff
@@ -0,0 +1,249 @@
...
+
+
+@value
+struct _EntityIterator[
+    archetype_mutability: Bool, //,
+    archetype_origin: Origin[archetype_mutability],
+    lock_origin: MutableOrigin,
+    *component_types: ComponentType,
+]:
+    """Iterator for over all entities corresponding to a mask.
```
mlange-42, line 99:
```suggestion
    """Iterator over all entities corresponding to a mask.
```

src/query.mojo
```diff
@@ -0,0 +1,249 @@
...
+            self._lock_ptr[].unlock(self._lock)
+        except Error:
+            debug_warn("Failed to unlock the lock. This should not happen.")
+
+    @always_inline
+    fn __iter__(owned self) -> Self as iterator:
+        iterator = self^
+
+    @always_inline
+    fn _find_archetypes(inout self):
```
mlange-42, line 197: Looks like `_EntityIterator` is a single-use throw-away thing. So I would try to avoid the `_archetype_indices` list and do the selection in `_next_archetype`.

src/world.mojo
```diff
@@ -279,6 +281,54 @@ struct World[*component_types: ComponentType]:
 
         return
 
+    @always_inline
+    fn get_entities(
```
mlange-42, line 285: Don't understand what this is. Just a query over all entities, irrespective of components? Isn't this covered by the method below?

src/query.mojo
samufi: This was my first idea as well. I decided against it because I wanted to know the size up front - and if I iterate through all archetypes anyway, I can also take a note of the archetypes of interest.
Knowing the size makes the `__has_next__` check simpler, which is called after prior to each step. 
Would you have an alternative solution or would you simply think that going through all archetypes is cheaper than allocating memory for the archetypes? Note that the iterator is only created once per query, so a few nanoseconds extra might be okay.

src/world.mojo
samufi: This is to prevent an issue with `InlineArray`, which has a constraint that its size must be above `0`. Hence, the program would fail to compile if only the function below existed and was called without a component. I will file an issue though and hope that they will remove the constraint so that this function can be removed.

src/query.mojo
mlange-42: I guess `__has_next__` is required for the iterator? This would of course complicate things.

> Would you have an alternative solution or would you simply think that going through all archetypes is cheaper than allocating memory

IMO these are not alternatives, but memory allocation is just on top. So you iterate all once anyway, either during query setup, or during the iteration. So either:
- Iterate all at init, allocate, iterate list during iteration, or
- Iterate all at iteration

> Note that the iterator is only created once per query

Sure, but each time you want to use the query. So basically, once per model step for each query. Given that (with Arche) query construction takes a few dozens of ns, this allocation is probably not insignificant.

mlange-42: BTW, using `while query.Next(): ...` would get rid of the need for `__has_next__`. I grant that iterators and for-loops are probably more ideomatic. Not sure about their overhead. Does Mojo make a zero-cost abstraction promise, and does it hold here?

samufi: It would be possible to make a query reusable, but this would lock the world until the reusable version is discarded. Maybe no good idea. 
Yes, the hasnext is needed for the for each iteration. We could maybe always cache the next archetype and use this for the check... I will have a look.

mlange-42: But a reusable query would require to populate the list again every time, as archetypes may have changed.

src/world.mojo
```diff
@@ -279,6 +281,54 @@ struct World[*component_types: ComponentType]:
...
+        """
+        return _EntityIterator(
+            self._component_manager,
+            Pointer.address_of(self._archetypes),
+            Pointer.address_of(self._locks),
+            BitMask(),
+        )
+
+    @always_inline
+    fn get_entities[
```
mlange-42, line 309: You should be aware that this syntax will only work if you stick to compile-time component reistration. If you want to allow runtime registration later, you will need an intermediate object like a filter, that users keep around and that does the ID lookup only once per run.

src/query.mojo
samufi: No, I meant to reuse exactly this list, which would require that nothing else could happen to the world in the mean time. I do not think this is a good idea. I have implemented a slicker version of query and I think this resolves the issue.

src/world.mojo
samufi: I think my next step is to write an example where I try out the ECS. If everything works fine as expected, I may lock in to the road of compile-time lookup, as this could have huge benefits at other sides as well. E.g., I would not need to store type sizes, and I might be able to save time passing IDs around.

src/query.mojo
mlange-42: Yeah, a query to keep around and have the world lock really makes no sense.

samufi: Just benchmarked the nuew version, and the query is much less efficient. Any ideas?

samufi: I think previously, the compiler was able to make some optimizations that are difficult now. I am compelled to revert to my first implementation, even if the query creation is a little longer. But factor 5 difference for each query iteration is a lot. Or would you say we should build a different test case?

src/world.mojo
mlange-42: Ok, that's weird. Maybe `__next__` is too complicated for inlining now? Or is `__len__` used by the iterator frequently?

According to the benchmarks, it looks to me like it is only slowed down when there are more then 1 component queried?

src/query.mojo
mlange-42: Ok, that's weird. Maybe `__next__` is too complicated for inlining now? Or is `__len__` used by the iterator frequently?

According to the benchmarks, it looks to me like it is only slowed down when there are more then 1 component queried?

mlange-42: Does Mojo have a profiling tool?

samufi: I dont't think they have a specialized profiler as of yet, but I think one could use one of the classical products as well. 
len is not used at all, I can comment it out. I think the problem has to do with having another branch to tackle. Inlining helps a lot, but still we have factor 2 worse performance. I don't understand why the number of components is so important here.

mlange-42: You could try to inline only the fast path, and move the inner `if` to a separate method.

> I don't understand why the number of components is so important here.

Yeah, that's why I think profiling would be helpful.
With just benchmarks, I would never have brought Arche to where is is now, profiling was really helpful during the optimization process.

mlange-42: Looks like you solved it?

samufi: It did not look like this on my own machine, but here it does. My last changes may have done the trick, though.

samufi: Is this ready to go now?

src/query.mojo
```diff
@@ -99,7 +99,7 @@ struct _EntityIterator[
 ]:
     """Iterator over all entities corresponding to a mask.
 
-    Locks the world while iterating.
+    Locks the world while it exisist.
```
mlange-42, line 102: Does that mean the world will only be unlocked when the iterator goes out of scope?

How do you get it out of scope if you need the world unlocked right after the iteration? Can you write anonymous blocks in Mojo? (Given Rust-like memory management, such blocks are a requirement IMO).

src/query.mojo
mlange-42: One more, due to your doc change. Wanted to ask for unlocking anyway...

src/query.mojo
samufi: https://docs.modular.com/mojo/manual/lifecycle/death/

mlange-42: I guess this also works when breaking out of the loop?

mlange-42 (approved):

src/query.mojo
samufi: Yes.

PR #31, opened 2025-01-08
Add an example for the usage of larecs

samufi:
The example features the simulation of the trajectory of satellites.

### DIFF
--- a/examples/satellites/magic.lock
+++ b/examples/satellites/magic.lock
@@ -1,0 +1,5567 @@
+version: 5
+environments:
+  default:
+    channels:
+    - url: https://conda.anaconda.org/conda-forge/
+    - url: https://conda.modular.com/max-nightly/
+    packages:
+      linux-64:
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
+      - conda: https://conda.anaconda.org/conda-forge/noarch/aiohappyeyeballs-2.4.4-pyhd8ed1ab_1.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/aiohttp-3.11.11-py312h178313f_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/aiosignal-1.3.2-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/alsa-lib-1.2.13-hb9d3cd8_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/annotated-types-0.7.0-pyhd8ed1ab_1.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/anyio-4.8.0-pyhd8ed1ab_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/noarch/attrs-24.3.0-pyh71513ae_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-auth-0.8.0-hb921021_15.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-cal-0.8.1-h1a47875_3.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-common-0.10.6-hb9d3cd8_0.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-compression-0.3.0-h4e1184b_5.conda
+      - conda: https://conda.anaconda.org/conda-forge/linux-64/aws-c-event-stream-0.5.0-h7959bf6_11.conda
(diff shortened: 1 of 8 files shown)


examples/satellites/systems.mojo
```diff
@@ -0,0 +1,63 @@
+from random import random
+from larecs import World
+from components import Position, Velocity
+from parameters import Parameters, GRAVITATIONAL_CONSTANT
+from python import PythonObject, Python
+
+
+fn move(mut world: World, parameters: Parameters) raises:
+    for properties in world.get_entities[Position, Velocity]():
```
mlange-42, line 9: IMO `properties` is quite confusing here. Better name it `entity`.

examples/satellites/systems.mojo
```diff
@@ -0,0 +1,63 @@
...
+        position = properties.get_ptr[Position]()
+        velocity = properties.get_ptr[Velocity]()
+
+        position[].x += velocity[].x * parameters.dt
+        position[].y += velocity[].y * parameters.dt
+
+
+fn accellerate(mut world: World, parameters: Parameters) raises:
```
mlange-42, line 17: This pattern will make it hard to come up with a `Systems` interface that is generally useful (in case you want to add systems to the library at all).

It is a good candidate for a [resource](https://mlange-42.github.io/arche/guide/resources/).

examples/satellites/model.mojo
```diff
@@ -0,0 +1,39 @@
+from larecs import World
+from parameters import Parameters
+from systems import move, accellerate, add_satellites, position_to_numpy
+from components import Position, Velocity
+from python import Python
+
+
+fn update(mut world: World, parameters: Parameters, time: Float64) raises:
```
mlange-42, line 8: `time` is also a good candidate for a resource. Systems may want to use it...

examples/satellites/magic.lock
```diff
@@ -0,0 +1,5567 @@
+version: 5
```
mlange-42, line 1: Does this file really need to be in the repo and under version control? 5k lines look like a hell lot if is will be in every little example.

examples/satellites/systems.mojo
samufi: I will change it, though I think that using the name entity can be confusing as discussed before.

examples/satellites/model.mojo
samufi: I rename it to `step`.

examples/satellites/systems.mojo
mlange-42: It is true that it is not really an entity. But for the code inside the loop, the user can treat it at an entity in the ODD (or even OOP) sense.

Generally, I would recommend to not try to invent your own terminology. Sticking to the established ECS terminology (query, entity, ...) will make it way easier for your users. For those fimiliar with ECS because they see just what they expect. For those new to ECS because they can relate the names in larecs with what they read about ECS in general.

README.md
```diff
@@ -51,11 +51,108 @@ To let VSCode and the LSP know of larecs, include it as follows:
...
+    var y: Float64
+
+@value
+struct Velocity:
+    var x: Float64
+    var y: Float64
+
+
+# Define a system
+fn move(mut world: World) raises:
```
mlange-42, line 78: I would work without systems, and without functions at all, in the README example.

examples/satellites/model.mojo
mlange-42: I mean, add this to the world as a resource once you have this feature. Don't make it a system argument.

README.md
samufi: That is, no move function? Everything in main?

mlange-42: Maybe to explain why...

With the approach of queries, systems are optional and just structure your code. So they are not a core feature that fits into the README example.

Other ECS implementation kind of "merge" systems and queries. I.e. by the definition of your system (e.g. by its generic types, or the function args), you define what you want to query (i'll give an example below). This has the IMO impractical effect that you can only ever iterate one "query" per system. Some even go as far as calling the system for every entity, not for the result of the query.

So, as systems are not required, they should not clutter the minimal example, and the example should not give users the (initial) impression that they are required.

mlange-42: > That is, no move function? Everything in main?

Yeah, exactly. Sorry, wrote the previous comment before your one arrived...

samufi: OK, updated the Readme. Good to go?

examples/satellites/magic.lock
mlange-42: Is it? Isn't that covered by the dependencies in the project's toml file? To my knowledge, such lock (or sum) files are there to prevent attacks by sneaking malware into a dependency.

Can you restruture later, so that there is only one `.lock` file covering all examples, instead of one for each?

samufi: I think I can restructure that way, yes. I needed an additional lock file only because the example has a python dependency for plotting that larecs should not have. Yes, the lock file has the purpose to ensure that the environment is exactly reproduced. The toml file provides constraints, the lock file the actual value if you will. I can say `magic update` inside the project and the lock file will be updated according to what is written in the toml.

mlange-42 (approved):

PR #43, opened 2025-01-17
Make Entity fields private

mlange-42:
Renames fields to `_id` and `_gen` to discourage manipulation.
Also hides `__init__` from the cocs using `@doc_private`.


samufi:
I think that private attributes should usually not be accessed outside of their class. I would suggest adding a getter for id. If this is done with `@always_inline`, there is no performance cost.

mlange-42:
> I think that private attributes should usually not be accessed outside of their class. I would suggest adding a getter for id. If this is done with `@always_inline`, there is no performance cost.

This still confronts users with something they should not bother with, but will do.

mlange-42:
> I think that private attributes should usually not be accessed outside of their class. I would suggest adding a getter for id. If this is done with `@always_inline`, there is no performance cost.

@samufi Done. Additionally, changed the type of `Entity.gen` to `UInt32`.

samufi:
Thanks! I typically prefer getters to have a `get_` prefix ("every method has a verb as name"). Should I change this myself?

samufi (approved):

mlange-42:
> Thanks! I typically prefer getters to have a `get_` prefix ("every method has a verb as name"). Should I change this myself?

Don't see the added benefit here, but as I will probably not convince you... did the renaming.

Off-topic: Can we please communicate using code comments? With plain comments, there are no threads and we have to quote-reply. Also, it is not possible to mark comments/threads as resolved.

PR #46, opened 2025-01-20
Set up doc-tests

mlange-42:
- adds a few testable code examples to the API docs
- sets up the CI to run doc-tests


docs/modo.yaml
```diff
@@ -0,0 +1,63 @@
...
+case-insensitive: false
+
+# Directories to scan for templates to overwrite builtin templates.
+templates: 
+  - modo
+
+# Shell commands to run before build as well as test.
+pre-run:
+  - |
+    set -e
```
mlange-42, line 38: @samufi This is required to stop and error in case of unsuccessful steps. Just like in the CI.

What do you think, should this be the default? I.e., should Modo add this internally before executing? Or are there cases where one would not want that?

Users could turn it of with `set +e`

samufi: How would modo behave on the command line in the two cases? I am not sure I have fully understood what e does.

mlange-42: Tests are only extracted if a test output path if given via config field or flag `tests`. Modo itself can't run the tests. Here, I added this as `post-test` step. pre- and post-steps can be skipped with flag `--bare`/`-B`:

```
modo build --bare
```

Regarding faulty docstrings, Modo gives warnings on things like unresolved refs, unclosed code blocks, ...
But there is also the flag `--strict`, which means to break with an error on any warnings.

samufi (approved):

PR #48, opened 2025-01-22
WIP: Introduce component manager as a parameter

samufi:
This means that size lookups can be done at compile time.


samufi:
Replaced by #49

Issue #27, opened 2025-01-02
Make the component manager a parameter


samufi:
Currently, the component manager contains static methods only. Therefore, it could also be used as a parameter. This could allow for faster (guaranteed compile-time) access of items and to implement iterators that are aware of the components that they include. This, in turn, would allow for better access functions and reduced overhead for `has` checks.

samufi:
Solved

PR #49, opened 2025-01-22
Introduce component manager as a parameter

samufi:
This improves performance and implements #27.


src/larecs/component.mojo
```diff
@@ -170,61 +165,20 @@ struct ComponentManager[*component_types: ComponentType]:
 
     @staticmethod
```
mlange-42, line 166: Looks a bit weird. But I guess each manager with a unique combination of components becomes a new type, so this can be static but different "instances" would still behave differently?

src/larecs/pool.mojo
```diff
@@ -18,6 +18,7 @@ struct EntityPool:
     var _next: EntityId
     var _available: Int
 
+    @always_inline
```
mlange-42, line 21: Is this really necessary? It will be called once in every run only.

src/larecs/pool.mojo
```diff
@@ -54,25 +56,30 @@ struct EntityPool:
         )
         self._available += 1
 
+    @always_inline
```
mlange-42, line 59: Same here.

src/larecs/pool.mojo
```diff
@@ -95,6 +102,7 @@ struct BitPool:
     # data type.
     var _length: UInt16
 
+    @always_inline
```
mlange-42, line 105: And again 😄

src/larecs/archetype.mojo
```diff
@@ -390,6 +383,7 @@ struct Archetype(CollectionElement, CollectionElementNew):
                 "Archetype does not contain component with id " + str(id) + "."
             )
 
+    @always_inline
```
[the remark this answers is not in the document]
samufi: The question is if inlining allows compiler optimizations that are otherwise not possible. I do not see inlining as a means to preventing call overhead but to getting better optimized code. The overall inlining commit improved the performance of create_entity from 15 ns to 7 ns.

My aproach to inlining is to leave all API functions as they are and to inline most of the internal stuff, which is often called a couple times only anyway.

I have not benchmarked every individual inlining step, though. If you think it is necessary, I am happy to do this.

src/larecs/component.mojo
samufi: Yes. This can be static because no local variables are accessed. Note that Mojo does not yet support class variables. My hope in declaring this as static is that it can be optimized more.

src/larecs/pool.mojo
samufi: Not necessary, but not an issue, either, since very simple. If it is called only once, there is also no cost associated with inlining and the potential advantage of optimizations remains.

src/larecs/pool.mojo
samufi: See discussion above.

src/larecs/pool.mojo
samufi: Response see above.

mlange-42 (approved):

PR #52, opened 2025-01-27
Collapse function and method docs

mlange-42:
Makes methods listings more concise by putting exerything bis the summary into collapsible `details` tags.

Uses Modo's template overwrite feature to add the `details` tag. Requires a Hugo shortcode `html` to allow for raw HTML in Markdown.

grafik

### DIFF
--- a/docs/modo/overload.md
+++ b/docs/modo/overload.md
@@ -1,0 +1,14 @@
+{{define "overload" -}}
+```mojo
+{{.Signature}}
+```
+
+{{`{{<html>}}`}}<details>
+<summary>{{`{{</html>}}`}}{{template "summary" . -}}{{`{{<html>}}`}}</summary>{{`{{</html>}}`}}
+{{template "description" . -}}
+{{template "func_parameters" . -}}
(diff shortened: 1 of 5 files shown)


docs/modo/overload.md
```diff
@@ -0,0 +1,14 @@
+{{define "overload" -}}
+```mojo
+{{.Signature}}
+```
+
+{{`{{<html>}}`}}<details>
```
samufi, line 6: Beautiful! If it is easy to do, a collapse/uncollapse all button could be nice.

samufi (approved):

docs/modo/overload.md
mlange-42: How about this? Or would you prefer text? There are tooltips at the moment.

grafik

samufi: I think text is better understandable. For tooltips, you'd need to hover.

mlange-42: Like this? 
grafik

samufi: Yes!

PR #66, opened 2025-02-05
Implement resource

samufi:
Implement resources for the world. These are handed over at world initialization.

Note that providing resources is a must, unless their type is explicitly specified as a parameter.

As a side, the archetype container was changed to a standard list. This reduces the code base that needs to be maintained. This was done during this commit to assert that an occurring issue was not due to the chained array list.

Known issue: capturing closures with World cause a segmentation fault. This is likely due to Mojo.


src/larecs/resource.mojo
```diff
@@ -0,0 +1,442 @@
...
+    var _initialized_flags: InlineArray[Bool, max(Self.size, 1)]
+
+    @always_inline
+    fn __init__(out self):
+        """Constructs an empty resource container."""
+
+        # Mark 'self.storage' as being initialized so we can work on it.
+        __mlir_op.`lit.ownership.mark_initialized`(
+            __get_mvalue_as_litref(self._storage)
+        )
```
mlange-42, line 159: Wow, this looks cryptic. Hope such stuff will be required only in very rare cases once Mojo is "finished".

src/larecs/world.mojo
```diff
@@ -16,10 +15,16 @@ from .component import (
...
+from .resource import ResourceType, ResourceContaining, Resources
 
 
 @value
-struct Replacer[mut: MutableOrigin, size: Int, *component_types: ComponentType]:
+struct Replacer[
+    mut: MutableOrigin,
+    size: Int,
+    *component_types: ComponentType,
+    resources_type: ResourceContaining,
```
mlange-42, line 26: Ah, ok... looks like this stuff is contiguous. You can't work with wildcards below?

mlange-42:
Looks like there are still outdated examples in some docstrings.

# 🧯🚀

src/larecs/resource.mojo
samufi: This is because I had to copy the implementation of Tuple, which was necessary in turn, because the origin did not match up otherwise. This is pure MLIR =).

mlange-42: Yeah, have seen this MLIR stuff already when reading up about Mojo.

mlange-42 (approved):

src/larecs/world.mojo
samufi: No, wildcards are a short form of parameterizing a method. They cannot be used in definitions.

PR #67, opened 2025-02-06
Use doc inheritance for `Resources`

mlange-42:


src/larecs/resource.mojo
```diff
@@ -288,34 +288,13 @@ struct Resources[*resource_types: ResourceType](ResourceContaining):
...
-        Parameters:
-            Ts: The Types of the resources to add.
-
-        Args:
-            resources: The resources to add.
-
-        Raises:
-            Error: If the resource already exists.
-        """
+        """See [[.ResourceContaining]]."""
```
samufi, line 291: Would we need to export ResourceContaining to make this work? Might be a good idea anyway (same as ComponentType, which I mistakingly removed last commit)

samufi: Yes, currently they are just an alias for CollectionElement. In the future, there may be a trait for "trivial" types that can be moved with a memcp. Components would need to satisfy this trait. At the moment, this is not enforced, which can lead to memory leaks and UB if users use non-trivial types. Therefore, I though it would be good to have this.

samufi: For resources, it may make sense to remove the trait...

samufi: The problem is not the compiler. The internals work with unsafe pointers. However, we cannot call the move constructors ore destructors of components until we can store / access the types in an Archetype at runtime. Currently this is only possible via parameters. That is, you can use stack-allocated memory, but you will have to free it yourself. 

There are some inline-data containers that could help, but they will always have a static maximal capacity.

mlange-42: > store / access the types in an Archetype

Wouldn't traits help here? As a very simple case, all components might have to implement a `Destructible` trait. Cast to that trait and call the destructor...

mlange-42: (ok, this would not cover move, which I guess is required for moving components between archetypes?)

samufi: > > store / access the types in an Archetype
> 
> Wouldn't traits help here? As a very simple case, all components might have to implement a `Destructible` trait. Cast to that trait and call the destructor...

You cannot cast to a trait afaik. What you envision is that the destructor is a virtual method that is looked up at runtime. (Even then it needs to be known at runtime) Such virtual methods do not exist (yet).

samufi: > (ok, this would not cover move, which I guess is required for moving components between archetypes?)

For move you could do the same as for destruction I think. But if the latter does not work....

mlange-42: can we resolve this?

samufi: But I do not think that using traits as suggested by you changes anything.

mlange-42: 😆 I rather meant: is your question answered and can we close this thread? You asked, so you should press the "resolve conversation" button when you are satisfied.

samufi:
A potential issues is that doc strings in the IDE do not know of the inheritance syntax. Would that be a reason to stick with redundant text for the moment?

mlange-42:
Shall we close this?

samufi:
> Shall we close this?

Maybe we should. I think the docs provided by the IDE are as important as the online docs. Sorry for requesting a feature that I don't use!

mlange-42:
> > Shall we close this?
> 
> Maybe we should. I think the docs provided by the IDE are as important as the online docs. Sorry for requesting a feature that I don't use!

No problem. Others may want to use it.

PR #68, opened 2025-02-07
Benchmark vs. AoS

mlange-42:
A bit of interpretation might also be useful. But as a first draft:

grafik


benchmark/aos_benchmark.mojo
```diff
@@ -0,0 +1,294 @@
...
+    # print(df)
+
+    fig_and_ax = plt.subplots(ncols=2, figsize=(10, 4))
+    fig = fig_and_ax[0]
+    ax = fig_and_ax[1]
+
+    ax1 = ax[0]
+    ax2 = ax[1]
+
+    for compExp in range(1, config.max_comp_exp, 1):
```
samufi, line 60: Let us use snake_case as general convention for variables.

benchmark/aos_benchmark.mojo
```diff
@@ -0,0 +1,294 @@
...
+            self.entities.append(AosEntity[Exp]())
+
+    @always_inline
+    fn update(mut self):
+        for entity in self.entities:
+            entity[].update()
+
+
+@value
+struct AosEntity[Exp: Int]:
```
samufi, line 228: Use capital letters for types and traits only. Maybe use `exponent` to avoid confusion with the `exp` function.

mojoproject.toml
```diff
@@ -10,3 +10,5 @@ version = "0.1.0"
 
 [dependencies]
 max = ">=25.1.0.dev2025011105"
+matplotlib = ">=3.0"
```
samufi, line 13: Do we really want a dependency of matplotlib in the main package? I would vote for creating a subpacke with its own dependencies. See the satellites example.

benchmark/aos_benchmark.mojo
```diff
@@ -0,0 +1,294 @@
...
+    var nanos_aos: Float64
+
+
+@value
+struct BenchConfig[max_comp_exp: Int]:
+    var max_entity_exp: Int
+    var target_iters: Int
+
+
+fn main() raises:
```
samufi, line 29: I would put main at the end of the file to conform to Python style.

benchmark/aos_benchmark.mojo
```diff
@@ -0,0 +1,294 @@
...
+
+    return BenchResult(
+        entities=entities,
+        components=2**Exp,
+        nanos_ecs=dur_ecs,
+        nanos_aos=dur_aos,
+    )
+
+
+fn createEcsWorld[Exp: Int](entities: Int) raises -> World:
```
samufi, line 195: Use an `out` argument rather than moving the result.

samufi (changes requested):

benchmark/aos_benchmark.mojo
mlange-42: Sure, it is just hard when switching between languages :D

benchmark/aos_benchmark.mojo
mlange-42: Aren't parameters/generics PascalCase in Mojo?

mojoproject.toml
mlange-42: No, definitely not. This is just interim. Actually, I just wanted to `pip` the dependencies in the CI. However, `matplotlib` was still not found, contrary to `pandas`.

benchmark/aos_benchmark.mojo
mlange-42: Thanks!

benchmark/aos_benchmark.mojo
mlange-42: changed to `component_exp`

benchmark/aos_benchmark.mojo
samufi: I know. But it still helps if you now where to search for the entry point. So I would go for the Python style even though it is not required,

benchmark/aos_benchmark.mojo
samufi: No, I think parameters are PascalCase only if they describe types (and not even always then). Tbh, I think Mojo is lacking a general style for parameters, and larecs is not fully consistent as well. I would go for lowercase params anyway if they are no types.

benchmark/aos_benchmark.mojo
mlange-42: Ok, will do. But is this really a set convention also for Mojo?

samufi: I don't know. Afaik there are no official conventions for Mojo as of yet.

mlange-42: moved `main` down

benchmark/plots/src/aos.mojo
```diff
@@ -0,0 +1,323 @@
+from collections import InlineArray
+from os import os
+from python import Python, PythonObject
+from time import perf_counter_ns
+
+import larecs as lx
+
+alias results_dir = "results"
```
samufi, line 8: use SCREAMING_SNAKE_CASE

benchmark/plots/src/aos.mojo
```diff
@@ -0,0 +1,323 @@
...
+"""Output directory for benchmark results."""
+
+alias target_iterations = 10**9
```
samufi, line 11: See above

benchmark/plots/src/aos.mojo
```diff
@@ -0,0 +1,323 @@
...
+    data["nanos_ecs"] = nanos_ecs
+    data["nanos_aos"] = nanos_aos
+
+    return pd.DataFrame(data)
+
+
+fn run_benchmarks(config: BenchConfig) raises -> List[BenchResult]:
+    results = List[BenchResult]()
+
+    for entExp in range(2, config.max_entity_exp, 1):
```
samufi, line 173: Is the step required? Remove otherwise.

benchmark/plots/src/aos.mojo
```diff
@@ -0,0 +1,323 @@
...
+        entities = 10**entExp
```
samufi, line 174: snake_case for entExp

benchmark/plots/src/aos.mojo
```diff
@@ -0,0 +1,323 @@
...
+        rounds = config.target_iters // entities
+        print(String(entities) + " entities")
```
samufi, line 176: Should work without explicit string conversion. Just do it comma separated and remove the space before " entitise".

benchmark/plots/src/aos.mojo
```diff
@@ -0,0 +1,323 @@
...
+            result = benchmark[compExp](rounds, entities)
+            results.append(result)
+
+    return results
+
+
+fn benchmark[
+    components_exp: Int
+](rounds: Int, entities: Int) raises -> BenchResult:
+    w1 = createEcsWorld[components_exp](entities)
```
samufi, line 189: createEcsWorld -> snake_case

benchmark/plots/src/aos.mojo
```diff
@@ -0,0 +1,323 @@
...
+    var start_ecs: Float64 = perf_counter_ns()
+    for _ in range(rounds):
+        for entity in w1.query[Position, Velocity]():
+            position = entity.get_ptr[Position]()
+            velocity = entity.get[Velocity]()
+            position[].x += velocity.x
```
samufi, line 195: Just so you know, in the benchmarking module there is also the `keep` method that prevents the compiler from optimizing away but has no extra cost.

benchmark/plots/src/aos.mojo
```diff
@@ -0,0 +1,323 @@
...
+
+    return BenchResult(
+        entities=entities,
+        components=2**components_exp,
+        nanos_ecs=dur_ecs,
+        nanos_aos=dur_aos,
+    )
+
+
+fn createEcsWorld[components_exp: Int](entities: Int, out w: World) raises:
```
samufi, line 213: snake_case for all methods.

mojoproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 authors = ["samufi <samufi@web.de>"]
 channels = ["conda-forge", "https://conda.modular.com/max-nightly"]
-description = "Add a short description here"
+description = "Larecs🌲 is a performance-oriented archetype-based ECS for Mojo🔥."
```
samufi, line 4: Thanks!

mojoproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
 authors = ["samufi <samufi@web.de>"]
```
samufi, line 2: Add yourself, too!

samufi (changes requested):

benchmark/plots/src/aos.mojo
mlange-42: Done

benchmark/plots/src/aos.mojo
mlange-42: Yeah, have seen this in your code. But don't think this is required here.

samufi: No, would just be a way to avoid the `+` operation, which is now part of the benchmark (which is fine, of course).

samufi: But I guess you wanted an assign operation. So stick with it.

mlange-42: I think this is a very classical example, so I would not try to get rid of the addition. Also, "write access" may make a difference, compared to just getting components for reading. Have seen this for component access through `world.get`.

samufi (approved):

PR #71, opened 2025-02-08
Implement Apply

samufi:
Apply allows to apply a closure to each component. This reduces not the need for manual iteration, but also allows the application of SIMD operations.

The advantages of SIMD seem to matter most when complex computations are performed (such as exp), so that the main time is not spent with memory management. Potentially, more efficient ways to handle the memory than currently applied in the benchmarks may exist.


src/larecs/test_utils.mojo
```diff
@@ -2,12 +2,56 @@ from testing import assert_true, assert_false, assert_equal
...
 from .world import World
 from .resource import Resources
 
 
+@always_inline
+fn load[
+    dType: DType, //, simd_width: Int, stride: Int = 1
+](ref val: SIMD[dType, 1], out simd: SIMD[dType, simd_width]):
+    """
+    Load multiple values from a SIMD.
```
mlange-42, line 17:
```suggestion
    Load multiple values from an SIMD.
```

mlange-42: Depends a bit on how this is spelled out "officially". Is it "an S.I.M.D", or "a sim-d."?

src/larecs/vectorize.mojo
mlange-42, line 1: Is this a leftover?

src/larecs/world.mojo
```diff
@@ -822,6 +823,76 @@ struct World[
...
+            accessor: MutableEntityAccessor
+        ) capturing -> None,
+        *Ts: ComponentType,
+        simd_width: Int = 1,
+        unroll_factor: Int = 1,
+    ](mut self) raises:
+        """
+        Applies an operation to all entities with the given components.
+
+        The operation is applied to chungs of `simd_width` entities,
```
mlange-42, line 860:
```suggestion
        The operation is applied to chunks of `simd_width` entities,
```

src/larecs/world.mojo
```diff
@@ -822,6 +823,76 @@ struct World[
...
+        unless not enough are available anymore. Then the chunk size is reduced.
+        The respecively used SIMD width is passed to the operation.
```
mlange-42, line 862: Don't see why "respectively" here.

src/larecs/world.mojo
```diff
@@ -822,6 +823,76 @@ struct World[
...
+
+        Caution! If `simd_width` is greater than 1, the operation **must**
+        apply to the `simd_width` elements after the element passed to
+        `operation`, assuming that the components are stored in contiguous
+        memory, respectively. This may require knowledge of the memory layout
```
mlange-42, line 867: Again, why "respectively"? Also, I don't understand what this wants to tell me as a user.

src/larecs/world.mojo
```diff
@@ -822,6 +823,76 @@ struct World[
...
+            operation(accessor)
+
+        self.apply[operation_wrapper, *Ts, unroll_factor=unroll_factor]()
+
+    fn apply[
+        operation: fn[simd_width: Int] (
+            accessor: MutableEntityAccessor
+        ) capturing -> None,
+        *Ts: ComponentType,
+        simd_width: Int = 1,
```
mlange-42, line 854: Doesn't this depend on the hardware? How can users know what to use here? How to deal with this when running on different hardware?

test/world_test.mojo
```diff
@@ -241,6 +242,78 @@ def test_world_reseource_access():
...
+
+    new_pos = pos.copy()
+    new_pos.x += vel.dx
+    new_pos.y += vel.dy
+
+    for _ in range(100):
+        _ = world.add_entity(pos, vel)
+
+    fn operation(accessor: MutableEntityAccessor) capturing:
+        try:
```
mlange-42, line 258: I quess this is required because operation can't be `raising` for some reason?

test/world_test.mojo
```diff
@@ -241,6 +242,78 @@ def test_world_reseource_access():
...
+        new_pos.x += vel.dx
+        new_pos.y += vel.dy
+        comparison.append(new_pos)
+
+    fn operation[simd_width: Int](accessor: MutableEntityAccessor) capturing:
+        try:
+            pos2 = accessor.get_ptr[Position]()
+            vel2 = accessor.get_ptr[Velocity]()
+
+            alias _load = load2[simd_width]
```
mlange-42, line 293: As `load` and `store` are in module `test_utils`, they seem not intended for library users. How are they intended to use `apply`?

src/larecs/world.mojo
mlange-42: Looking at the tests, I gen an idea now. An example would probably be helpful here.

src/larecs/test_utils.mojo
samufi: I would go with Sim-D, as this is the way people at Mojo talk about it. So, I would not change it to `an SIMD`.

src/larecs/vectorize.mojo
samufi: Yes. I'll remove it.

src/larecs/world.mojo
samufi: Changing in new commit,

src/larecs/world.mojo
samufi: To make clear that the SIMD width used is not always the `simd_width` passed to the `apply`function. Maybe I should formulate this differently.

src/larecs/world.mojo
samufi: I will add an example.

src/larecs/world.mojo
samufi: You can use an arbitrary SIMD width, but you can also use `simdwidthof`. In my experience, doubling this value led to better results, though...

test/world_test.mojo
samufi: Yes, this is because I use `vectorize` internally, which cannot handle raising functions. I could handle the errors myself, but only after the loop finished. This does not seem to be clean to me, so I kept it unraising, knowing that this is inconvenient.

src/larecs/world.mojo
```diff
@@ -822,6 +823,76 @@ struct World[
         if not self._entity_pool.is_alive(entity):
             raise Error("The considered entity does not exist anymore.")
 
+    @always_inline
+    fn apply[
```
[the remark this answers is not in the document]
samufi: I would guess that the aliases would always be "unrolled". For example, the operation is actually defined via an alias `fn[simd_width: Int] (accessor: MutableEntityAccessor) capturing -> None`, but this is transformed to this very long cryptic thing, because `MutableEntityAccessor` is an alias for something long and cryptic.

test/world_test.mojo
mlange-42: But what if the operation raises an error? Would that go unnoticed given the empty `except`?

test/world_test.mojo
mlange-42: I think if it is covered by an example that does not use the `test_utils` stuff, it is ok to leave this to users.

samufi: I have added such an example to the API docs. Adding this stuff to the main docs is a todo.

src/larecs/world.mojo
mlange-42: What happens when the given width exceeds the size of the registers?

samufi: I do not know what happens on the technical level, but in practice this is just as if more data is loaded to memory and then processed iteratively.

test/world_test.mojo
mlange-42: User code could still raise, independent of the ECS.

Would there be a way for the user to handle errors more useful than just printing? Or would there be only `sys.exit` (or similar)? E.g., can we see a backtrace without being able to raise?

Not that I think you should (or even could) change anything in this implementation. Just interested...

src/larecs/world.mojo
samufi: I think the only disadvantage of using a too large SIMD width is loosing memory locality to some extent.

test/world_test.mojo
samufi: Yes, users could define a variable outside of the closure (if they are careful not to use copy capturing) and save the error to it (or if errors are no structs, simply set a boolean to True) and raise it after application of vectorie.

mlange-42 (approved):

test/world_test.mojo
samufi: Example:

```python

from collections import Optional
def test_world_apply_raises():
    world = SmallWorld()
    pos = Position(1.0, 2.0)
    vel = Velocity(0.1, 0.2)

    new_pos = pos.copy()
    new_pos.x += vel.dx
    new_pos.y += vel.dy

    for _ in range(100):
        _ = world.add_entity(pos, vel)

    error = Optional[Error]()
    
    @parameter
    fn operation(accessor: MutableEntityAccessor) capturing:
        if not error:
            try:
                raise Error("Test")
            except e:
                error = e

    world.apply[operation, Position, Velocity, unroll_factor=3]()

    if error:
        raise error.value()
```

samufi: I could offer to implement this pattern directly in Larecs, too. Unfortunately, there seems to be no stack trace printed. This may be an issue with Mojo, though.

mlange-42: > I could offer to implement this pattern directly in Larecs

That would be `apply` raising the error?

samufi: Yes.

PR #73, opened 2025-02-11
Implement `World.query.without` to exclude components

mlange-42:


src/larecs/query.mojo
```diff
@@ -51,17 +160,14 @@ struct _EntityIterator[
         archetypes: Pointer[List[Self.Archetype], archetype_origin],
         lock_ptr: Pointer[LockMask, lock_origin],
         owned mask: BitMask,
+        owned without_mask: Optional[BitMask],
```
samufi, line 163: Use an overload of init here. Along the lines of the following:

```python

    fn __init__(
        out self,
        archetypes: Pointer[List[Self.Archetype], archetype_origin],
        lock_ptr: Pointer[LockMask, lock_origin],
        owned mask: BitMask,
    ) raises:
        constrained[
            not Self.has_without_mask,
            "No without_mask provided",
        ]()
        self = Self.__init__[check_has_without_mask=False](
            archetypes,
            lock_ptr,
            mask^,
            BitMask(),
        )

    fn __init__[check_has_without_mask: Bool = True](
        out self,
        archetypes: Pointer[List[Self.Archetype], archetype_origin],
        lock_ptr: Pointer[LockMask, lock_origin],
        owned mask: BitMask,
        owned without_mask: BitMask,
    ) raises:
        @parameter
        if check_has_without_mask:
            constrained[
                Self.has_without_mask,
                "has_without_mask is False, but a without_mask was provided.",
            ]()
         ...
```

test/query_test.mojo
```diff
@@ -20,65 +20,77 @@ def test_query_length():
         _ = world.add_entity(c1, c2, c3)
         _ = world.add_entity(c0, c1, c2, c3)
 
-    assert_equal(len(world.query[FlexibleComponent[0]]()), 4 * n)
-    assert_equal(len(world.query[FlexibleComponent[1]]()), 4 * n)
-    assert_equal(len(world.query[FlexibleComponent[2]]()), 4 * n)
-    assert_equal(len(world.query[FlexibleComponent[3]]()), 4 * n)
+    assert_equal(len(world.query[FlexibleComponent[0]]().__iter__()), 4 * n)
```
samufi, line 23: Create a (trivial) len implementation for Query that does this.

samufi (changes requested):

src/larecs/query.mojo
samufi: Oh, instead of `if if check_has_without_mask` include that into the constraint directly. More concise.

src/larecs/query.mojo
```diff
@@ -135,8 +239,13 @@ struct _EntityIterator[
             self._archetype_index_buffer[self._buffer_index] + 1,
             self._archetype_count,
         ):
+            mask = self._archetypes[][i].get_mask()
```
samufi, line 242: Use an `@parameter if` here. Maybe introduce the `mask` variable only in case where you need it. It makes a copy of a bitmask.

samufi:
You could also update the Readme's coming soon section...

src/larecs/query.mojo
```diff
@@ -71,6 +177,7 @@ struct _EntityIterator[
         self._lock = self._lock_ptr[].lock()
         self._archetype_count = len(self._archetypes[])
         self._mask = mask^
+        self._without_mask = without_mask
```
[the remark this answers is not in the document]
mlange-42: done

src/larecs/query.mojo
mlange-42: removed the temporary variable

src/larecs/query.mojo
```diff
@@ -9,6 +9,114 @@ from .lock import LockMask
...
+            mask: The mask of the components to iterate over.
+        """
+        self._archetypes = archetypes
+        self._lock_ptr = lock_ptr
+        self._mask = mask^
+        self._without_mask = None
+
+    @always_inline
+    fn __iter__(
+        owned self,
```
samufi, line 68: I am wondering if this makes a copy of self. Maybe better not owned?

src/larecs/query.mojo
samufi: Well, in the @parameter case it may save you a lookup. Not sure if this significant, though.

src/larecs/query.mojo
mlange-42: It is not owned anymore, as we want to be able to call this multiple times.

src/larecs/query.mojo
```diff
@@ -51,26 +186,46 @@ struct _EntityIterator[
...
         """
+
+        if Self.has_without_mask == (without_mask is None):
+            raise Error(
+                "has_without_mask and without_mask do not match.",
+                Self.has_without_mask,
+                without_mask is None,
+            )
+
+        # TODO: replace above check by constraint.
```
mlange-42, line 209: Was not able to get this to work. In the currently commented-out code, it seems like the (run-time) conditions are ignored and constraints are always applied, and thus fail.

samufi: Again, look at the example I provided regarding the overloaded inits.

samufi (changes requested):

src/larecs/query.mojo
```diff
@@ -1,21 +1,154 @@
...
+        """
+        self._world = world
+        self._mask = mask^
+        self._without_mask = None
+
+    fn __len__(self) raises -> Int:
+        """
+        Returns the number of entities remaining in the iterator.
+
+        Note that this requires the creation of an iterator ftom the query.
```
samufi, line 65: Typo

src/larecs/query.mojo
```diff
@@ -1,21 +1,154 @@
...
+        """
+        Creates an iterator over all entities that match the query.
+
+        Returns:
+            An iterator over all entities that match the query.
+
+        Raises:
+            Error: If the lock cannot be acquired (more than 256 locks exist).
+        """
+        iterator = self._world[]._get_iterator[has_without_mask](
```
samufi, line 91: Better do this via an overload on `_get_iterator` and an `@parameter if` here

src/larecs/query.mojo
```diff
@@ -119,25 +275,34 @@ struct _EntityIterator[
...
+        Fills the _archetype_index_buffer with the
         archetypes' indices.
         """
         buffer_index = 0
         for i in range(
             self._archetype_index_buffer[self._buffer_index] + 1,
             self._archetype_count,
         ):
+
+            @parameter
```
samufi, line 291: Just put the `@parameter if` as an additional layer of the conditional clause below. Do not store an intermediate variable.

src/larecs/world.mojo
```diff
@@ -1072,10 +1073,37 @@ struct World[
...
+        iterator = Query[
+            __origin_of(self),
+            *component_types,
+            resources_type=resources_type,
+        ](
+            Pointer.address_of(self),
+            BitMask(Self.component_manager.get_id_arr[*Ts]()),
+        )
+
+    fn _get_iterator[
```
samufi, line 1085: Use an overload as mentioned above.

test/query_test.mojo
```diff
@@ -203,6 +205,113 @@ def test_query_has_component():
         assert_false(entity.has[FlexibleComponent[3]]())
 
 
+fn test_query_empty() raises:
+    world = SmallWorld()
+    query = world.query[FlexibleComponent[0]]()
+    cnt = 0
```
samufi, line 211: cnt?

samufi (changes requested):

src/larecs/query.mojo
mlange-42: ok, now that worked

src/larecs/query.mojo
```diff
@@ -53,24 +188,68 @@ struct _EntityIterator[
...
             lock_ptr: a pointer to the world's locks.
             mask: The mask of the components to iterate over.
+            without_mask: The mask for components to exclude.
 
         Raises:
             Error: If the lock cannot be acquired (more than 256 locks exist).
         """
+
+        @parameter
+        if check_has_without_mask:
```
samufi, line 236: If you do it this way, then the name `check_has_without_mask` is misleading, as it suggests that a check is only done if it is true. I do not think the second constraint, because it could also be triggered if a without_mask was given.

src/larecs/query.mojo
```diff
@@ -51,26 +186,46 @@ struct _EntityIterator[
...
+        #        not Self.has_without_mask,
+        #        "has_without_mask is False, but a without_mask was provided.",
+        #    ]()
+
         self._archetypes = archetypes
         self._lock_ptr = lock_ptr
         self._lock = self._lock_ptr[].lock()
         self._archetype_count = len(self._archetypes[])
         self._mask = mask^
+        self._without_mask = without_mask^.or_else(BitMask())
```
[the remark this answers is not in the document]
mlange-42: Outdated (or rather, moved to world), but tests indicate that is works. Why should it not?

src/larecs/query.mojo
mlange-42: Really? You bother about creating a bool variable?

samufi: I bother about an additional branch.

samufi: `@parameter if` does not open a new branch.

test/query_test.mojo
mlange-42: You mean I should rename this to `counter`?

src/larecs/query.mojo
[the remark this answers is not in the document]
samufi: Because you first move it and then do the or_else. But maybe this makes perfect sense.

test/query_test.mojo
samufi: Yes.

src/larecs/query.mojo
```diff
@@ -1,21 +1,151 @@
...
+            f[] += 1
+        ```
+
+        Parameters:
+            Ts: The types of the components to exclude.
+
+        Returns:
+            The query, exclusing the given components.
+        """
+        result = Query[
```
samufi, line 127: And you can use Self.Query here as well.

src/larecs/query.mojo
mlange-42: Renamed to `has_without_mask`. But may be a little bit confusing now, with `has_without_mask` and `Self.has_without_mask`.

Not sure what you mean about the escond constraint. Simplified the entire thing to this:

```mojo
constrained[
    has_without_mask == Self.has_without_mask,
    "has_without_mask and presence of without_mask don't match.",
]()
```

src/larecs/query.mojo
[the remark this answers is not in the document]
mlange-42: Yeah, the `Optional` is owned, not its content I would think.

src/larecs/query.mojo
samufi: My suggestion is to have a parameter for enforcing that the parameter `Self.has_without_mask` is True, as the init obviously takes a mask. This constraint should only be lifted in the special case where a query without a without_mask is created. I find this more logical than the new solution proposed by you.

src/larecs/query.mojo
samufi: Yes.

mlange-42: creates ugly code duplication as I can't put compile-time and run-time stuff into the same condition, but well...

samufi: Actually, I am not sure if a nested if might be less efficient, because the parameter if adds another runtime if, which may be an additional branch...

src/larecs/query.mojo
mlange-42: Not sure why contradicting parameters should be allowed, but done.

src/larecs/query.mojo
mlange-42: then i don't understand how you want this to look like. in the more simple case of len, it looks like this now:

```mojo
@parameter
if has_without_mask:
    if (
        not self._archetypes[]
        .unsafe_get(i)
        .get_mask()
        .contains_any(self._without_mask)
    ):
        size += len(self._archetypes[][i])
else:
    size += len(self._archetypes[][i])
```

I can't move the mask check into the comp-time if, as it is dynamic.

src/larecs/query.mojo
```diff
@@ -1,21 +1,202 @@
...
+    resources_type: ResourceContaining,
+    has_without_mask: Bool,
+]:
+    """Query builder for entities with and without specific components."""
+
+    alias Query = Query[
+        world_origin,
+        *component_types,
+        resources_type=resources_type,
+        has_without_mask=_,
```
samufi, line 25: Not needed. If not given, a wildcard is assumed.

src/larecs/query.mojo
```diff
@@ -1,21 +1,202 @@
...
+    ]
+
+    alias Iterator = _EntityIterator[
+        _,
+        _,
+        *component_types,
+        component_manager = ComponentManager[*component_types](),
+        has_without_mask=_,
```
samufi, line 33: Same here

src/larecs/query.mojo
```diff
@@ -1,21 +1,202 @@
...
+    ]
+
+    var _world: Pointer[
+        World[*component_types, resources_type=resources_type], world_origin
```
samufi, line 37: Use an alias for World. Later you can always use Self.World.component_manager

src/larecs/query.mojo
```diff
@@ -1,21 +1,202 @@
...
+        self._without_mask = BitMask()
+
+    fn __init__(
+        out self,
+        world: Pointer[
+            World[*component_types, resources_type=resources_type], world_origin
+        ],
+        owned mask: BitMask,
+        owned without_mask: BitMask,
+    ) raises:
```
samufi, line 87: Remove raises in init, if it does not raise. constraints do not raise.

samufi: Same in other init

src/larecs/query.mojo
```diff
@@ -1,21 +1,202 @@
...
+            f[] += 1
+        ```
+
+        Parameters:
+            Ts: The types of the components to exclude.
+
+        Returns:
+            The query, exclusing the given components.
+        """
+        result = Self.Query[has_without_mask=True](
```
samufi, line 186: If I see this right, has_without_mask is always True when you use this alias. So, you can define it that way right away. And you could reflect that in its name. QueryWithWithout

src/larecs/query.mojo
```diff
@@ -53,24 +236,63 @@ struct _EntityIterator[
...
+        ]()
+        self = Self.__init__[False](
+            archetypes,
+            lock_ptr,
+            mask^,
+            BitMask(),
+        )
+
+    fn __init__[
+        has_without_mask: Bool
```
samufi, line 262: Give it a True default. Then the developer does not need to care about this anywhere except in the case above.

src/larecs/query.mojo
```diff
@@ -53,24 +236,63 @@ struct _EntityIterator[
...
             lock_ptr: a pointer to the world's locks.
             mask: The mask of the components to iterate over.
+            without_mask: The mask for components to exclude.
 
         Raises:
             Error: If the lock cannot be acquired (more than 256 locks exist).
         """
+
+        @parameter
+        if has_without_mask:
```
samufi, line 284: You can put both parameters together into the constraint. So, no if, but constrained[has_without_mask and check_has_without_mask]

src/larecs/query.mojo
samufi, line 262: Make the parameter "check" the consistenty. (or "enforce") So rename it to `check_has_without_mask`.

src/larecs/world.mojo
```diff
@@ -1026,11 +1026,11 @@ struct World[
     @always_inline
     fn query(
```
samufi, line 1027: You may also want to use an alias here

samufi (changes requested):

src/larecs/query.mojo
mlange-42: Sure? But then, both must be true.

samufi: I think in my mental model, the parameter means the opposite of what you have right now. check_has_without should be True by default, but when calling the function from the overload, it should be False

src/larecs/query.mojo
samufi: Sorry for messing around. It's not intentional.

src/larecs/query.mojo
mlange-42: As a query can only return a QueryWithWithout, I donot really see the added benefit of the ugly name. But done.

src/larecs/query.mojo
mlange-42: And the (altered) constraind would fail then.

mlange-42: So I changed this to `check_...`. But still, it does not make sense put it into the constraint, as this would mean it always has to be true.

mlange-42: what would work is `not check or has` in the constraint

mlange-42: Done.

src/larecs/world.mojo
```diff
@@ -1042,24 +1043,15 @@ struct World[
         Raises:
             Error: If the world is [.World.is_locked locked].
         """
-        iterator = _EntityIterator(
-            Pointer.address_of(self._archetypes),
-            Pointer.address_of(self._locks),
+        iterator = Query[has_without_mask=False](
```
samufi, line 1046: Or `Self.Query(...)`

src/larecs/world.mojo
```diff
@@ -1072,10 +1064,51 @@ struct World[
         Raises:
             Error: If the world is [.World.is_locked locked].
         """
-        iterator = _EntityIterator(
+        iterator = Query[has_without_mask=False](
```
samufi, line 1067: See above.

src/larecs/world.mojo
```diff
@@ -1072,10 +1064,51 @@ struct World[
...
+        has_without_mask: Bool
+    ](
+        mut self,
+        owned mask: BitMask,
+        out iterator: _EntityIterator[
+            __origin_of(self._archetypes),
+            __origin_of(self._locks),
+            *component_types,
+            component_manager = Self.component_manager,
+            has_without_mask=has_without_mask,
```
samufi, line 1082: Let's use consistent spacing around =. I thought the formatter would do that...

src/larecs/world.mojo
```diff
@@ -1072,10 +1064,51 @@ struct World[
...
+    ](
+        mut self,
+        owned mask: BitMask,
+        owned without_mask: Optional[BitMask],
+        out iterator: _EntityIterator[
+            __origin_of(self._archetypes),
+            __origin_of(self._locks),
+            *component_types,
+            component_manager = Self.component_manager,
+            has_without_mask=has_without_mask,
```
samufi, line 1102: Spacing around =

src/larecs/world.mojo
```diff
@@ -1026,12 +1032,7 @@ struct World[
     @always_inline
     fn query(
         mut self,
-        out iterator: _EntityIterator[
-            __origin_of(self._archetypes),
-            __origin_of(self._locks),
-            *component_types,
-            component_manager = Self.component_manager,
-        ],
+        out iterator: Self.Query[__origin_of(self)],
```
samufi, line 1035: the output is not an iterator anymore. Adjust name to, e.g., `query`.

src/larecs/world.mojo
```diff
@@ -1072,10 +1064,51 @@ struct World[
...
             Error: If the world is [.World.is_locked locked].
         """
-        iterator = _EntityIterator(
+        iterator = Query[has_without_mask=False](
+            Pointer.address_of(self),
+            BitMask(Self.component_manager.get_id_arr[*Ts]()),
+        )
+
+    fn _get_iterator[
+        has_without_mask: Bool
```
samufi, line 1073: Make this False by default, since a different value should never be used.

src/larecs/world.mojo
```diff
@@ -1072,10 +1064,51 @@ struct World[
...
+    ) raises:
+        iterator = _EntityIterator[has_without_mask=has_without_mask](
             Pointer.address_of(self._archetypes),
             Pointer.address_of(self._locks),
-            BitMask(Self.component_manager.get_id_arr[*Ts]()),
+            mask,
+        )
+
+    fn _get_iterator[
+        has_without_mask: Bool
```
samufi, line 1092: Make this True by default

samufi (changes requested):

src/larecs/world.mojo
mlange-42: Th formatter did what you see. It adds spaces for `Self...` stuff, and removes them otherwise. So I don't know what I can do about it.

samufi: Then so be it.

src/larecs/world.mojo
mlange-42: see above

src/larecs/query.mojo
```diff
@@ -1,21 +1,202 @@
...
+    var _mask: BitMask
+    var _without_mask: BitMask
+
+    fn __init__(
+        out self,
+        world: Pointer[
+            World[*component_types, resources_type=resources_type], world_origin
+        ],
+        owned mask: BitMask,
+    ) raises:
```
samufi: This does not raise, does it?

src/larecs/query.mojo
samufi: Does not raise?

src/larecs/query.mojo
```diff
@@ -1,21 +1,197 @@
...
+    alias World = World[*component_types, resources_type=resources_type]
+
+    alias QueryWithWithout = Query[
+        world_origin,
+        *component_types,
+        resources_type=resources_type,
+        has_without_mask=True,
+    ]
+
+    alias Iterator = _EntityIterator[
```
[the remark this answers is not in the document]
mlange-42: Done.

src/larecs/query.mojo
mlange-42: It is `self._lock_ptr[].lock()` in the iterator constructor that raises and needs to bubble up all the way. But I will check if there are any `raises` that can still be removed.

mlange-42: Could remove from `world.query` as well as Query methods, except `__iter__`.

Not sure how you could comment on the Query constructors, as they don't raise. Maybe you accidentially used an outdated commit?

src/larecs/query.mojo
mlange-42: No, but is also did not raise 4h ago. Outdated commit?

src/larecs/query.mojo
```diff
@@ -53,24 +236,63 @@ struct _EntityIterator[
...
+            "No without_mask provided",
+        ]()
+        self = Self.__init__[False](
+            archetypes,
+            lock_ptr,
+            mask^,
+            BitMask(),
+        )
+
+    fn __init__[
```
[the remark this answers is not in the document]
mlange-42: Yes, due to `self._lock_ptr[].lock()`.

samufi (approved):

PR #76, opened 2025-02-12
Rework satellites example to use resources

mlange-42:


examples/satellites/model.mojo
```diff
@@ -3,28 +3,29 @@ from parameters import Parameters
...
 
 fn main() raises:
-    world = World[Position, Velocity](Resources())
+    world = World[Position, Velocity](Resources[Parameters]())
 
-    parameters = Parameters(dt=0.1, mass=5.972e24)
+    world.resources.add(Parameters(dt=0.1, mass=5.972e24))
 
-    add_satellites(world, 300)
+    add_satellites(world, 50)
```
samufi, line 20: Why less satellites?

examples/satellites/model.mojo
```diff
@@ -3,28 +3,29 @@ from parameters import Parameters
...
 from python import Python
+from sys import argv
 
 
-fn update(mut world: World, parameters: Parameters, step: Float64) raises:
-    for _ in range(Int(step / parameters.dt)):
-        move(world, parameters)
-        accellerate(world, parameters)
+fn update(mut world: World, step: Float64) raises:
+    for _ in range(Int(step / world.resources.get_ptr[Parameters]()[].dt)):
```
samufi, line 10: Is it better style to add every parameter as its individual resource or to use a struct like here?

samufi (approved):

examples/satellites/model.mojo
mlange-42: I prefer an intermediate solution, grouping closely related parameters. Take a look at [beecs](https://github.com/mlange-42/beecs/blob/main/params/default.go#L52) for an example (sub-structs are added as individual resources).

examples/satellites/model.mojo
mlange-42: Leftover from testing, as it was not clear whether compilation or the run took so long. Will fix...

examples/satellites/model.mojo
mlange-42: As this are only two parameters, I would leave it like this. It is sufficient for demonstration.

mlange-42:
@samufi Benchmarks are crashing, which can't be the source of this MR.

samufi:
> @samufi Did you make a back-merge here? Why that? In most cases this just makes things messy, better make a rebase.

I did (pushed a button on GH). Won't do it again.

samufi:
> @samufi Benchmarks are crashing, which can't be the source of this MR.

Please rebase on current PR after merge. This should fix it.

PR #78, opened 2025-02-13
Improve Query docs

mlange-42:
* move examples from constructors to struct docstring
* fix docs on `__len__`

### DIFF
--- a/src/larecs/query.mojo
+++ b/src/larecs/query.mojo
@@ -16,7 +16,24 @@ struct Query[
     resources_type: ResourceContaining,
     has_without_mask: Bool,
 ]:
-    """Query builder for entities with and without specific components."""
+    """Query builder for entities with and without specific components.
+
+    This type not be used directly, but through the [..world.World.query] method:
+
+    ```mojo {doctest="query_init" global=true hide=true}
+    from larecs import World, Resources, MutableEntityAccessor
+    ```
+
+    ```mojo {doctest="query_init"}
+    world = World[Float64, Float32, Int](Resources())
+    _ = world.add_entity(Float64(1.0), Float32(2.0), 3)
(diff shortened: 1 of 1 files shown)


src/larecs/query.mojo
```diff
@@ -16,7 +16,24 @@ struct Query[
     resources_type: ResourceContaining,
     has_without_mask: Bool,
 ]:
-    """Query builder for entities with and without specific components."""
+    """Query builder for entities with and without specific components.
+
+    This type not be used directly, but through the [..world.World.query] method:
```
samufi, line 21: The type should...

src/larecs/query.mojo
```diff
@@ -107,7 +96,7 @@ struct Query[
 
     fn __len__(self) raises -> Int:
         """
-        Returns the number of entities remaining in the iterator.
+        Returns the number of entities matching the query.
```
samufi, line 99: If this is still the original implementation, the length decreases while the iterator is used up.

samufi: Hence, the new comment may not be fully accurate.

samufi (changes requested):

src/larecs/query.mojo
mlange-42: oops

samufi: should not of course

samufi: The verb was missing

src/larecs/query.mojo
mlange-42: We are in `Query` here. A new iterator is created below, so the statemant should be correct IMO.

samufi: Oh. Convincing.

samufi (approved):

PR #83, opened 2025-03-14
Set the groundwork for batch operations

samufi:
- Add necessary data structures for batch operations. 
- Implement batch creation of entities
- Simplify entity creation

### DIFF
--- a/src/larecs/comptime_optional.mojo
+++ b/src/larecs/comptime_optional.mojo
@@ -1,0 +1,127 @@
+from memory import UnsafePointer
+
+
+@value
+struct ComptimeOptional[
+    ElementType: CollectionElement,
+    has_value: Bool = True,
+](Movable, Copyable, ExplicitlyCopyable):
+    """An optional type that can potentially hold a value of ElementType.
+
+    In contrast to the built-in oprtional, it is decided at
+    compile time whether a value is present. This allows
+    this type to have a size of 0 in these cases.
+
+    Parameters:
+        ElementType: The type of the elements in the array.
+        has_value: Whether the element exists.
+    """
+
+    # Fields
+    alias type = __mlir_type[
+        `!pop.array<`, Int(has_value).value, `, `, Self.ElementType, `>`
+    ]
+    var _value: Self.type
+    """The underlying storage for the optional."""
+
+    # ===------------------------------------------------------------------===#
+    # Life cycle methods
+    # ===------------------------------------------------------------------===#
+
+    @always_inline
+    fn __init__(out self):
+        """This constructor will always cause a compile time error if used.
+        It is used to steer users away from uninitialized memory.
+        """
+        constrained[
+            not has_value,
+            "Initialize with a value if `has_value` is `True`",
+        ]()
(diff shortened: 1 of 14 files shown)


src/larecs/comptime_optional.mojo
```diff
@@ -0,0 +1,133 @@
...
+    var _value: Self.type
+    """The underlying storage for the optional."""
+
+    # ===------------------------------------------------------------------===#
+    # Life cycle methods
+    # ===------------------------------------------------------------------===#
+
+    @always_inline
+    @implicit
+    fn __init__(out self, none: None = None):
```
mlange-42, line 33: Why is the unused argument required?

src/larecs/comptime_optional.mojo
```diff
@@ -0,0 +1,133 @@
...
+    @implicit
+    fn __init__(out self, owned value: Self.ElementType):
+        """Constructs an optional type holding the provided value.
+
+        Args:
+            value: The value to fill the optional with.
+        """
+        __mlir_op.`lit.ownership.mark_initialized`(__get_mvalue_as_litref(self))
+
+        var ptr = UnsafePointer.address_of(self.value())
```
mlange-42, line 53: Does this really need to use unsafe? Can't you just store the value in `_value`?

src/larecs/world.mojo
```diff
@@ -270,46 +271,113 @@ struct World[
...
+        #         newRel = &arch.RelationComponent
+
+        #     var bits = subscription(true, false, len(comps) > 0, false, newRel != nil, newRel != nil)
+        #     var trigger = self._listener.Subscriptions() & bits
+        #     if trigger != 0 && subscribes(trigger, &arch.Mask, nil, self._listener.Components(), nil, newRel):
+        #         self._listener.Notify(self, EntityEventEntity: entity, Added: arch.Mask, AddedIDs: comps, NewRelation: newRel, EventTypes: bits)
+
+        return
 
     fn add_entity[
```
mlange-42, line 363: Wouldn't it make more sense to use plural for the method name?

src/larecs/comptime_optional.mojo
samufi: For the implicit constructor (`@implicit`). This does that whenever the type is required but only the constructor arguments are given the type is constructed with these arguments.

Example:
```
fn fun(a: ComptimeOptional[Int, True]):
   pass

fun(1) # works but requires @implicit
fun(ComptimeOptional[Int, True](1)) # This would be required without the @implicit 
```

The `None` argument makes it possible to provide `None` instead of an empty `ComptimeOptional`.

src/larecs/world.mojo
samufi: Yes, this would make sense. Would this be your preference? (This was my question on MM; I am undecided.)

mlange-42: Yes, I would prefer plural, let alone for reasoning when reading the code.

samufi: Done.

mlange-42 (approved):

PR #84, opened 2025-03-16
Simplify the iterator creation in world by removing overloads

samufi:
Using the new ComptimeOptional type, the code can be strongly simplified.

### DIFF
--- a/src/larecs/comptime_optional.mojo
+++ b/src/larecs/comptime_optional.mojo
@@ -110,6 +110,23 @@ struct ComptimeOptional[
         return UnsafePointer(ptr)[]
 
     @always_inline
+    fn or_else(self, value: ElementType) -> ElementType:
+        """Returns a copy of the value contained in the Optional or a default value if no value is not present.
+
+        Args:
+            value: The dedault value to return if the optional is empty.
+
+        Returns:
+            A copy of the value contained in the Optional or the default value if no value is present.
+        """
+
+        @parameter
+        if has_value:
+            return self.value()
+        else:
+            return value
+
+    @always_inline
     fn unsafe_ptr(self) -> UnsafePointer[Self.ElementType]:
         """Get an `UnsafePointer` to the underlying array.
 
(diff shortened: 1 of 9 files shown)


src/larecs/comptime_optional.mojo
```diff
@@ -109,6 +109,23 @@ struct ComptimeOptional[
         )
         return UnsafePointer(ptr)[]
 
+    @always_inline
+    fn or_else(self, value: ElementType) -> ElementType:
```
mlange-42, line 113: Just as a note, you may want to take a look how Rust does this.

In Rust, this would be `or`, while `or_else` takes a callback. So the argument to `or` is always evaluated. Contrary, `or_else` evaluates the argument only in the else-case, which makes sense when constructing the argument has an associated cost.

src/larecs/comptime_optional.mojo
```diff
@@ -109,6 +109,23 @@ struct ComptimeOptional[
...
+        """Returns a copy of the value contained in the Optional or a default value if no value is not present.
```
mlange-42, line 114: Double negation.

src/larecs/comptime_optional.mojo
```diff
@@ -109,6 +109,23 @@ struct ComptimeOptional[
...
+
+        Args:
+            value: The dedault value to return if the optional is empty.
```
mlange-42, line 117: Typo in "default"

src/larecs/query.mojo
```diff
@@ -10,12 +10,14 @@ from .resource import ResourceContaining
 from .debug_utils import debug_warn
 from .comptime_optional import ComptimeOptional
 
+from benchmark import keep
```
mlange-42, line 13: Not required anymore I guess.

src/larecs/comptime_optional.mojo
samufi: Done.

src/larecs/comptime_optional.mojo
samufi: Fixed.

src/larecs/query.mojo
samufi: Fixed.

PR #86, opened 2025-03-21
Implement batch removal of entities

samufi:


src/larecs/query.mojo
```diff
@@ -162,6 +164,39 @@ struct Query[
         )
 
 
+@value
+struct QueryInfo[
```
mlange-42, line 168: Why is this required? Can't you just use a simple query? Or maybe hide this from the user and just get it inside batch methods?

samufi: This is required for accepting Query as an argument in a method of world. Query holds a reference to world. This leads to the multiple references access the same memory error. Hence, I need to extract the two masks from the query. To do this implicitly, I use this struct. The user does not need to care about this, but I do not see a way to not use it in the function signature.

mlange-42: Ok, already suspected something like this. Have seen the implicit stuff in the docstring. Just not so nice that this has to be exposed to the user, and thus also should be in the API refs. But it is how it is.

PR #90, opened 2025-03-28
Flexible resources

samufi:
Implement flexible resources that can be added at runtime.

### DIFF
--- a/src/larecs/component.mojo
+++ b/src/larecs/component.mojo
@@ -46,7 +46,7 @@ fn get_sizes[
 
 
 @always_inline
-fn _contains_type[T: ComponentType, *Ts: ComponentType]() -> Bool:
+fn contains_type[T: ComponentType, *Ts: ComponentType]() -> Bool:
     @parameter
     for i in range(len(VariadicList(Ts))):
 
@@ -120,7 +120,7 @@ struct ComponentManager[
 
         # This constraint will fail if the component type is not in the list.
         constrained[
-            _contains_type[T, *component_types](),
+            contains_type[T, *component_types](),
             "The used component is not in the component parameter list.",
         ]()
 
--- a/src/larecs/component.mojo
+++ b/src/larecs/component.mojo
@@ -90,7 +90,7 @@ struct ComponentManager[
     alias component_count = len(VariadicList(component_types))
     alias component_sizes = get_sizes[*component_types]()
 
-    fn __init__(mut self):
+    fn __init__(out self):
         constrained[
             dType.is_integral(),
             "dType needs to be an integral type.",
--- a/src/larecs/resource.mojo
+++ b/src/larecs/resource.mojo
@@ -88,8 +88,8 @@ struct Resources[TypeMap: TypeMapping = DynamicTypeMap]:
         """Adds resources.
 
         Parameters:
-            M: The type of the mapping from resources to IDs.
             Ts: The Types of the resources to add.
+            M: The type of the mapping from resources to IDs.
 
         Args:
             resources: The resources to add.
@@ -151,8 +151,8 @@ struct Resources[TypeMap: TypeMapping = DynamicTypeMap]:
         """Sets the values of resources.
 
         Parameters:
-            M: The type of the mapping from resources to IDs.
             Ts: The types of the resources to set.
+            M: The type of the mapping from resources to IDs.
             add_if_not_found: If true, adds resources that do not exist.
 
         Args:
@@ -221,8 +221,8 @@ struct Resources[TypeMap: TypeMapping = DynamicTypeMap]:
         """Removes resources.
 
         Parameters:
-            M: The type of the mapping from resources to IDs.
             Ts: The types of the resources to remove.
+            M: The type of the mapping from resources to IDs.
 
         Raises:
             Error: If one of the resources does not exist.
@@ -271,8 +271,8 @@ struct Resources[TypeMap: TypeMapping = DynamicTypeMap]:
         """Gets a resource.
 
         Parameters:
-            M: The type of the mapping from resources to IDs.
             T: The type of the resource to get.
+            M: The type of the mapping from resources to IDs.
 
         Returns:
             A reference to the resource.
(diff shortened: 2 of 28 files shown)


test/world_test.mojo
```diff
@@ -295,31 +296,52 @@ def test_remove_and_add():
...
     var value: Int
 
 
 @value
 struct Resource2:
+    alias id = 2
     var value: Int
 
 
 def test_world_reseource_access():
```
mlange-42, line 309: typo

test/world_test.mojo
```diff
@@ -295,31 +296,52 @@ def test_remove_and_add():
 
 @value
 struct Resource1:
+    alias id = 1
```
mlange-42, line 299: I would consider this rather dangerous. Did the hash-based solution not work, or did you remove it for simplicity?

src/larecs/resource.mojo
```diff
@@ -327,109 +135,211 @@ struct Resources[*resource_types: CollectionElement](ResourceContaining):
...
+
+    fn set[
         *Ts: CollectionElement,
-        raise_if_found: Bool = False,
-        raise_if_not_found: Bool = False,
-    ](
-        mut self, owned resources: VariadicPack[_, CollectionElement, *Ts]
-    ) raises:
-        """Sets the resources.
+        M: StaticallyTypeMapping,
```
mlange-42, line 148: IMO this should be either `StaticTypeMap` or `StaticallyTypedMap`, just for the grammar. Probably the former.

src/larecs/type_map.mojo
```diff
@@ -0,0 +1,51 @@
...
+    fn __init__(out self):
+        """Initializes the type mapping."""
+        ...
+
+
+trait IdentifiableCollectionElement(CollectionElement):
+    alias id: IdType
+
+
+trait StaticallyTypeMapping(TypeMapping):
```
mlange-42, line 16: Would use `Static` instead of `Statically`, for proper grammar.

src/larecs/resource.mojo
```diff
@@ -246,75 +54,75 @@ struct Resources[*resource_types: CollectionElement](ResourceContaining):
...
+        return len(self._storage)
 
-    @always_inline
-    fn _unsafe_get_ptr[
-        idx: Int
-    ](ref self) -> UnsafePointer[resource_types[idx.value]]:
-        """Gets an unsafe pointer to a resource.
+    fn add[
+        *Ts: IdentifiableCollectionElement
+    ](mut self: Resources[DynamicTypeMap], owned *resources: *Ts) raises:
```
mlange-42, line 68: Personally, I would not consider the var-args particularly useful here (i.e. adding multiple per call). Same for the other methods. But this is just a matter of preference (for simplicity).

src/larecs/component.mojo
```diff
@@ -46,7 +46,7 @@ fn get_sizes[
 
 
 @always_inline
-fn _contains_type[T: ComponentType, *Ts: ComponentType]() -> Bool:
+fn contains_type[T: ComponentType, *Ts: ComponentType]() -> Bool:
```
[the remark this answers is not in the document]
samufi: Because this can be useful always and does not need to be internal.

src/larecs/resource.mojo
samufi: "It is statically type mapping". "Type mapping" ist das adjektiv und das wird von "statically" näher spezifiziert. Vielleicht aber auch "StaticlyTypeMapping"... 
Ich werd das sowieso löschen...

src/larecs/type_map.mojo
samufi: See above.

src/larecs/resource.mojo
samufi: I understand that the current implementation is way too verbose. I think I like the general implementation, as it costs nothing...

test/world_test.mojo
samufi: Fixed.

src/larecs/type_map.mojo
mlange-42: Does not sound correct for me, but you have more experience in EN language.

test/world_test.mojo
samufi: Can you remind me of the hash based solution?

mlange-42: This: https://github.com/samufi/larecs/pull/81/files#diff-1f45aa0cdfe48d3345ded3e9c68016743257ba15320a1a042399c24fe8186bbfR18

samufi: OK, updated the PR. Now good? Please re-approve.

src/larecs/type_map.mojo
```diff
@@ -1,22 +1,144 @@
...
+    struct MyStruct(IdentifiableCollectionElement):
+        alias id = TypeId("my_package.my_module.MyStruct")
+    ```
+
+    If the ID is used in a limited context only, you can
+    set the `id` manually by passing an integer to the constructor.
+    """
+
+    var _id: UInt
+    var _name: StringLiteral
```
mlange-42, line 35: For the case of hash collisions, or why do you want to store this?

samufi: Yes. And the overhead is minimal if I understand string literals correctly.

src/larecs/type_map.mojo
```diff
@@ -1,22 +1,144 @@
...
+    fn __eq__(self, other: Self) -> Bool:
+        """Checks if two IDs are equal.
+
+        Args:
+            other: The other ID to compare with.
+
+        Returns:
+            True if the IDs are equal, False otherwise.
+        """
+        return self._id == other._id and self._name == other._name
```
mlange-42, line 74: This might slow down dict lookups. But it could be that equality is checked only in case of hash collisions, not sure.

samufi: The comparison may be executed once per lookup to confirm that there is no hash collision. I think this okay here, as it is not performance-critical.

mlange-42 (approved):

PR #93, opened 2025-04-01
Write a user guide

samufi:


docs/src/guide/installation.md
```diff
@@ -0,0 +1,35 @@
...
+2. Navigate to the `src/` subfolder.
+3. Execute `mojo package larecs`.
+4. Move the newly created file `larecs.mojopkg` to your project's source directory.
+
+## Include the source directly when compiling
+
+Instead of building Larecs🌲 as a package, you can also include its
+source code directly when running or compiling your own project.
+This has the advantage that you can access the source while debugging 
+and (test-wise) adjust the Larecs🌲 source code.
```
mlange-42, line 20:
```suggestion
and (test-wise) adjusting the Larecs🌲 source code. 
```

(because of "debugging"). Would probably skip the "test-wise".

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,105 @@
++++
+type = "docs"
+title = "Entities, components, and the world"
+weight = 10
++++
+
+## A little bit of introduction...
```
mlange-42, line 7: Would not start chapters with a level 2 heading, but with just a bit of introductory text. Particularly in this case, but also for the other chapters. It also looks a bit weird.

grafik

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,105 @@
...
+
+Larecs🌲 provides an efficient way to store, access and manipulate 
+data belonging to computational entities.
```
mlange-42, line 10: "computational" makes no sense here for me.

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,105 @@
...
+"Entities" often represent objects or agents that have certain
+properties ("objects" in the object oriented framework),
+here called "Components". Which attributes an entity has 
+can change at runtime, making the approach extremely flexible.
+Furthermore, entities do not belong to pre-defined 
+"classes" but are rather fully characterized by their attributes.
+A key functionality of entity component systems is to 
+query all entities that have certain components and process them
+in a corresponding way.
```
mlange-42, line 19: This entire first section is probably quite confusing for readers not familiar with ECS.

IMO the entire page would be more comprehensible if it is just structured like the title:

- Intro (no heading): 1-3 sentences about ECS (in general)
- Section "Entities":
  - What they represent: Game objects, "ODD entities", or potentially anything else that "lives" in the game or sim
  - What they are: IDs to access components (link next section)
- Section "Components":
  - What they represent: Entity data / state variables of entities
  - What they are: structs (or rather, any unique type?); (current) limitation trivial ...
- Section "The world"
  - central storage for all this stuff
  - entities are bound to a world
  -  ...

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,105 @@
...
+    var x: Float64
+    var y: Float64
+
+@value
+struct Velocity:
+    var dx: Float64
+    var dy: Float64
+```
+
+In this example, we consider a two-dimensional world.
```
mlange-42, line 55: Unnecessary detail that just extends the text, IMO. It is obvious.

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,105 @@
...
+to look up the entity's components. As a consequence,
+entities are small in memory and can be efficiently 
+stored, and their data can only be accessed via the world.
+
+> [!NOTE] 
+> Though entities can safely be stored an passed around,
+> their components (or pointers to them) should never be stored
+> externally, as they can move in memory at any time.
+
+How entities are created and used will be discussed in the
```
mlange-42, line 104: Would always use present for cross refs: "is discussed".

You shouls also think about how to name "parts" of the docs. I use "chapter" for pages and "section" for sections (i.e. headings) of a chapter. Here, I would not know how to call sections, as it is already reserved.

docs/src/guide/entities_components_world.md
samufi: What do we call entities if they have not been defined? How do we explain entity without using the word?

mlange-42: I think the entire chapter requires restructuring, which would resolve the issue. See [below](https://github.com/samufi/larecs/pull/93#discussion_r2023714128).

docs/src/guide/installation.md
samufi: Fixed.

docs/src/guide/installation.md
```diff
@@ -0,0 +1,35 @@
...
+You can include the Larecs🌲 source code as follows:
+
+```
+mojo run -I "path/to/larecs/src" example.mojo
+```
+
+## Include Larecs🌲 in VSCode and its language server
+
+To let VSCode and its language server know of Larecs🌲 (so that not all occurrences
+of larecs get red underlines), include it as follows:
```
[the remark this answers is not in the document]
samufi: Done.

docs/src/guide/entities_components_world.md
samufi: Okay, will do.

docs/src/guide/entities_components_world.md
samufi: Sure, but this was a genuine question: How do you explain entity without using the word?

docs/src/guide/entities_components_world.md
samufi: Removed.

docs/src/guide/entities_components_world.md
samufi: Ok, will do.

docs/src/guide/entities_components_world.md
mlange-42: Use the word entity, sure. Just "*computational* entity" makes no sense for me. Maybe for an HPC cluster or something, but not for ECS entities. What would that be, a "rechnerische Entität"?

docs/src/guide/entities_components_world.md
samufi: Please have a look at the updated version.

docs/src/guide/entities_components_world.md
samufi: Please have a look at the updated version

mlange-42:
@samufi I need to run the update script to be able to run `modo`. This gives me uncommitted changes to the `.lock` files. Is this intentional?

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Entities
+
+Entities are the central unit around which the data in
+an ECS is organized. Entities could represent game objects,
+or potentially anything else that "lives" in a game or simulation.
+Each entity can possess an arbitrary
+set of components, which can also be used to characterize
+an entity. That is, in an ECS, cars would not directly 
+identified as "cars" but rather as everything that has
```
mlange-42, line 28:
```suggestion
be identified as "cars" but rather as something that has 
```

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,111 @@
...
+    var dx: Float64
+    var dy: Float64
+```
+
+These structs are not only used to store data. Their
+types are also used as unique identifiers for the components
+of an entity. That is, they have a similar role to 
+a key in a dictionary.
+
+How components can be associated with entities is described
```
mlange-42, line 73: IMO always writing "associated" makes things harder to understand. Although components are not stored in entities, I would prefer "can be added to entities".

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,111 @@
...
+    var dy: Float64
+```
+
+These structs are not only used to store data. Their
+types are also used as unique identifiers for the components
+of an entity. That is, they have a similar role to 
+a key in a dictionary.
+
+How components can be associated with entities is described 
+in the [later chapters](../adding_and_removing_entities).
```
mlange-42, line 74: Why not use concrete chapter titles?

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,111 @@
...
+> components in Larecs🌲. That is, structs that can be
+> copied and moved via simple memory operations. This does
+> *not* include structs that manage heap-allocated memory
+> such as `List` or `Dict`. Typically, it is not advisable to
+> use such objects in the ECS context anyway; however, 
+> Larecs🌲 might support such "complex" structs in a future version. 
+
+## The world
+
+The central container type of Larecs🌲 is the
```
mlange-42, line 87:
```suggestion
The central container of Larecs🌲 is the
```

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,111 @@
...
+{{< api World >}}. The `World` stores all data and information about
```
mlange-42, line 88:
```suggestion
{{< api World >}}. It stores all data and information about
```

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,111 @@
...
+the state of the entities and their surroundings.
```
mlange-42, line 89: Surroundings? This is probably very confising.

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,111 @@
...
+The `World` struct provides the main functionality
+of the ECS, such as adding and removing entities,
+looking up components, and iterating over entities.
+
+Larecs🌲 gains a lot of its efficiency by using compile-time
```
mlange-42, line 94: I do not agree here. Actually, I think it gains only the last small bit of performance from it, and the majority from the general architecture. But as you really seem to like this fact, go ahead... 😆

docs/src/guide/changing_entities.md
```diff
@@ -0,0 +1,137 @@
++++
+type = "docs"
+title = "Changing entities"
+weight = 30
++++
+
+The values of an entity's component can be 
+accessed and changed via the {{< api World.get get >}} and {{< api World.get_ptr get_ptr >}}
+methods of world. Here, `get` returns a reference, 
+which can be changed in-place only and `get_ptr`
```
mlange-42, line 10: For me, "in-place" would not be clear, and it would not convey what you probably want to say (that it does not change the actual component inside the storage).

docs/src/guide/changing_entities.md
```diff
@@ -0,0 +1,137 @@
...
+    var dx: Float64
+    var dy: Float64
+```
+
+```mojo {doctest="guide_change_entities" hide=true}
+world = World[Position, Velocity]()
+```
+
+```mojo {doctest="guide_change_entities"}
+# Add an entity and get its representation
```
mlange-42, line 34:
```suggestion
# Add an entity with a component
```

docs/src/guide/changing_entities.md
```diff
@@ -0,0 +1,137 @@
...
+entity = world.add_entity(Position(0, 0))
+
+# Get a copy of the position
+pos = world.get[Position](entity)
+
+# This does not change the entity's state!
+pos.x = 5 
+print(world.get[Position](entity).x == pos.x) # False
+
+# Changing the reference in-place works, though.
```
mlange-42, line 44: Ah, ok, mow I get it. You see, even knowing all these internals, I misunderstood the above text.

docs/src/guide/changing_entities.md
```diff
@@ -0,0 +1,137 @@
...
+world.get[Position](entity).x = 5
+
+# Similarly, replacing the component completely 
+# works as well.
+world.get[Position](entity) = Position(5, 0)
+
+# For later access of the position, use a pointer
```
mlange-42, line 51: This ("later access") seemingly contradics the warning on component pointers in the previous section, although I know what you mean.

docs/src/guide/changing_entities.md
```diff
@@ -0,0 +1,137 @@
...
+pos_ptr = world.get_ptr[Position](entity)
+
+# Now, changing the value via the local pointer variable works.
+pos_ptr[].x = 10 # Use the `[]` operator to dereference the pointer
+print(world.get[Position](entity).x == 10) # True
```
mlange-42, line 56: I would tend to split all this up into multiple examples and plain text.

docs/src/guide/changing_entities.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+Of course, accessing a component only works if the entity has
+the component in question. Accessing a component 
+that the entity does not have, will result in an error.
```
mlange-42, line 61:
```suggestion
that the entity does not have will result in an error.
```

docs/src/guide/changing_entities.md
```diff
@@ -0,0 +1,137 @@
...
+# Check if the entity has a velocity component
+if world.has[Velocity](entity):
+    print("Entity has a velocity component")
+else:
+    print("Entity does not have a velocity component")
+```
+
+## Setting multiple components at once
+
+We can set multiple components at once using the {{< api World.set set >}}`
```
mlange-42, line 85:
```suggestion
We can set multiple components at once using the {{< api World.set set >}} 
```

docs/src/guide/changing_entities.md
```diff
@@ -0,0 +1,137 @@
...
+if world.has[Velocity](entity):
+    print("Entity has a velocity component")
+else:
+    print("Entity does not have a velocity component")
+```
+
+## Setting multiple components at once
+
+We can set multiple components at once using the {{< api World.set set >}}` 
+method. This method takes a tuple of components and sets them
```
mlange-42, line 86: This is not a tuple. It just takes one or more components.

docs/src/guide/changing_entities.md
```diff
@@ -0,0 +1,137 @@
...
+We can set multiple components at once using the {{< api World.set set >}}` 
+method. This method takes a tuple of components and sets them 
+all in one go.
+
+```mojo {doctest="guide_change_entities"}
+# Add an entity with two components
+entity = world.add_entity(Position(0, 0), Velocity(1, 1))
+
+# Set multiple components at once
+world.set(entity, Position(5, 5), Velocity(2, 2))
```
mlange-42, line 94: I guess this is only possible for components that the entity already has. This is not clear here.

Further, I would tend to move this after the section on adding and removing.

Woudl als replace "We can set multiple components..." by "We can set the values of multiple components...".

docs/src/guide/changing_entities.md
```diff
@@ -0,0 +1,137 @@
...
+world.replace[Position]().by(Velocity(2, 2), entity=entity)
+```
+
+Similar to the `add` and `remove` 
+methods, this works with arbitrary numbers of
+components, so we can replace any number of components with
+any other number of new components.
+
+> [!Tip]
+> Replacing components way in one go is significantly
```
mlange-42, line 135:
```suggestion
> Replacing components in one go is significantly 
```

You shoudl also add that adding/removing multiple is better than one by one. It is logical, but may not be completely obvious.

docs/src/guide/entities_components_world.md
mlange-42: The general structure and comprehensibility is much better now IMO. However, I still do not really like the introductory paragraph. Will try to come up with a suggestion tomorrow.

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,111 @@
...
+around individual entities. Suppose we wanted to 
+model a world with different objects and agents that each have 
+certain properties. The idea of entity component systems 
+(ECS) is to characterize these objects and agents 
+(called "entities" in an ECS) via the set of 
+attributes ("components") they possess, and to 
+model the interactions between them via "systems",
+functions that operate on entities 
+with certain components. Larecs🌲 provides the functionality 
+and tools to implement such entity component systems.
```
mlange-42, line 18: To be exact, it is not "an entity component system", i.e. it is not a system, but "entity component system architecture". If you want to call something the "system", than that's Larecs, not the thing you build with it.

So if you want to include such a statement, it should be something like "Larecs provides the functionality to build [something] following the ECS architecture".

But TBH, neither "Larecs🌲 provides a flexible and efficient way to store..." nor "Larecs🌲 provides the functionality..." are required here IMO. In my eyes, that's just fillers, as it is clear that we are talking about Larecs here.

docs/src/guide/entities_components_world.md
samufi: Done.

docs/src/guide/entities_components_world.md
samufi: Because it is multiple chapters. Would you want me to list them all? I mainly want to tell the reader "hold on for now, it will become clear".

docs/src/guide/entities_components_world.md
samufi: Container "structure"? Otherwise one might think that larecs is contained...

samufi: Or "the central data structure"

docs/src/guide/entities_components_world.md
samufi: Done.

docs/src/guide/entities_components_world.md
samufi: I would like to write "the rest of the system", but "system" is a reserved word...

docs/src/guide/entities_components_world.md
samufi: "Larecs🌲 gains efficiency and usability by exploiting
Mojo's compile-time programming capabilities."?

docs/src/guide/changing_entities.md
samufi: Well it does change the component inside the storage if you use in-place operations.
```pyhton
world.get[Position]().x = 3 # works
world.get[Position]() = Position(0, 3) # works

a = world.get[Position]() 
a = 5 #does not work.
```

docs/src/guide/changing_entities.md
samufi: Done.

docs/src/guide/changing_entities.md
samufi: What about this?

Here, `get` returns a reference, 
which becomes a copy of the component if stored in a local variable,
and `get_ptr` returns a pointer, which can write into  
the original memory even if stored locally.

docs/src/guide/changing_entities.md
samufi: `# To access the position later in this method, use a pointer`

docs/src/guide/changing_entities.md
samufi: Fixed

docs/src/guide/changing_entities.md
samufi: Fixed

docs/src/guide/changing_entities.md
samufi: Fixed

docs/src/guide/changing_entities.md
samufi: Done

docs/src/guide/entities_components_world.md
samufi: I await your suggestion for the intro section. I am very happy that you improve it!

docs/src/guide/changing_entities.md
samufi: > I guess this is only possible for components that the entity already has.

You can control that behaviour via a parameter. I did not want to go into that level of detail here, but will do if you think it is helpful.

> Further, I would tend to move this after the section on adding and removing.

I find the current place more logical... Do you have *strong* feelings about this?

> "We can set the values of multiple components..."

Done.

docs/src/guide/changing_entities.md
samufi: I have split it up into two sections for references and pointers.

docs/src/guide/entities_components_world.md
mlange-42: Ok, then maybe "following chapters"?

docs/src/guide/entities_components_world.md
mlange-42: Maybe then "... stores entities, components and resources" (with a link on the latter)?
Or "... entities and their associated data"?

docs/src/guide/changing_entities.md
mlange-42: Well, the last "works", but has no effect.

docs/src/guide/changing_entities.md
mlange-42: > I find the current place more logical... Do you have strong feelings about this?

I would expect this to be rarely used. Firstly, in most cases users will get a pointer and manipulate fields. Secondly, it is (in both cases) just an alternative to "pointer assignment" (``get_ptr() = ...`) and `add`, resp.

mlange-42:
> Throw away the update results and run the above-mentioned command chain.

Ok, the `mojo doc` part works now. Bit for the `mojo test` part I get this error:

```
Included from /tmp/test-8e2d07.mojo:12:
/root/projects/larecs/docs/test/larecs/larecs_world_World_add_entity_test.mojo:5:18: error: invalid initialization: expected at most 0 positional arguments, got 1
    world = World(Resources())
            ~~~~~^~~~~~~~~~~~~
Included from /root/projects/larecs/src/larecs/__init__.mojo:1:
Included from /root/projects/larecs/src/larecs/__init__.mojo:82:
/root/projects/larecs/src/larecs/world.mojo:154:8: note: function declared here
    fn __init__(
       ^
Included from /tmp/test-8e2d07.mojo:12:
/root/projects/larecs/docs/test/larecs/larecs_world_World_add_entity_test.mojo:6:9: error: use of unknown declaration 'world'
    e = world.add_entity()
        ^~~~~
/root/projects/larecs/.magic/envs/default/bin/mojo: error: failed to parse the provided Mojo source module
/root/projects/larecs/.magic/envs/default/bin/mojo: error: couldn't build the test executable. Exit code 1
Error: in script post-test: exit status 1
```

docs/src/guide/adding_and_removing_entities.md
```diff
@@ -0,0 +1,100 @@
...
+```mojo {doctest="guide_add_remove_entities"}
+# Add a batch of 10 entities with given position and velocity
+x_position = 0
+for entity in world.add_entities(Position(0, 0), Velocity(1, 0), count=10):
+    entity.get[Position]().x = x_position
+    x_position += 1
+```
+
+More information on manipulation of and iteration over entities 
+will be provided in the upcoming chapters.
```
mlange-42, line 74: Would use presence. It is already there...

docs/src/guide/adding_and_removing_entities.md
```diff
@@ -0,0 +1,100 @@
...
+
+> [!Note]
+> Iterators block certain changes to the world and should not
+> be stored in a variable. That is, use the result of `add_entities` 
+> only in the right hand side of for loop.
```
mlange-42, line 79: Either `of a for loop` or `of for loops`.

If I would not know from your detailing or good knowledge of Mojo, I would ask myself what happens when not using the iterator.

I think the world lock even deserves a separate section, e.g explaining what is blocked, and how to handle this limitation. You woll need a similar tip in the chapter on queries.

docs/src/guide/entities_components_world.md
samufi: Done.

docs/src/guide/entities_components_world.md
samufi: Hmm, resources is difficult to understand before it has been introduced. I would like to find a different term for that. "entities and associated data" would sound to me like that each entity has associated data (which then would be components).

docs/src/guide/adding_and_removing_entities.md
samufi: Fixed.

docs/src/guide/adding_and_removing_entities.md
samufi: > Either of a for loop or of for loops.

Fixed.

> I think the world lock even deserves a separate section

Ok, I will add a section to the queries chapter

> You woll need a similar tip in the chapter on queries.

That is already present

docs/src/guide/_index.md
```diff
@@ -3,4 +3,14 @@ type = "docs"
 title = "User guide"
 weight = -1
 +++
```
mlange-42, line 5: You could unfold the user guide sections by adding this to the preface:

```toml
[sidebar]
open = true
```

(or similar, not 100% sure about the toml syntax)

docs/src/guide/queries_iteration.md
```diff
@@ -0,0 +1,292 @@
...
+
+```mojo {doctest="guide_queries_iteration"}
+# Add entities with different components
+_ = world.add_entity()
+_ = world.add_entity(Position(0, 0))
+_ = world.add_entity(Velocity(1, 0))
+_ = world.add_entity(Position(1, 0), Velocity(1, 0))
+
+# Query all entities that have a position
+query = world.query[Position]()
```
mlange-42, line 50: Didn't you write earlier that iterators should not be stored in variables? Might be confusing.

docs/src/guide/queries_iteration.md
```diff
@@ -0,0 +1,292 @@
...
+conforming to a query.
+
+## Queries
+
+{{< api Query Queries >}} allow to iterate over all 
+entities with or without a specific 
+set of components. To create a query, we can use 
+the {{< api World.query query >}} method of 
+{{< api World >}}. The parameters used in this method are the components
+that each entity we look for must have. For example, if we want to
```
mlange-42, line 20: It might be useful to mention (here or later) that this is non-exclusive.

docs/src/guide/queries_iteration.md
samufi: Yes, but queries are no iterators.

samufi: Nonetheless, indeed, queries hold a pointer to world, which causes ownership issues making the program fail to compile.

docs/src/guide/queries_iteration.md
```diff
@@ -0,0 +1,292 @@
...
+exactly the components we are looking for but no more.
+This can be done using the {{< api Query.exclusive exclusive >}} 
+method. For example, if we want to iterate
+over all entities that have only a `Position` component,
+we can do this as follows:
+
+```mojo {doctest="guide_queries_iteration"}
+excluding_query = world.query[Position]().exclusive()
+print(len(excluding_query)) # "1"
+```
```
mlange-42, line 86: All queries in these examples use only a single component. Maybe add an example that uses more than one, to prevent the expression that only single components can be queried?

docs/src/guide/queries_iteration.md
```diff
@@ -0,0 +1,292 @@
...
+
+> [!Note] 
+> Determining the length of a query is not a trivial operation
+> and can be expensive if the ECS is involves many components.
+> Therefore, it is advisable to avoid applying the `len` function
+> on queries in "hot" code.
```
mlange-42, line 92: In the Ark user guide, I mentioned that it is still way faster than using iteration for counting.

docs/src/guide/queries_iteration.md
```diff
@@ -0,0 +1,292 @@
...
+
+## Iterating over queries
+
+As we have seen, we can iterate over queries using a for loop.
+Here, the control variable ("entity") is an {{< api EntityAccessor >}} 
+object, i.e., not technically an {{< api Entity >}}, which is
+merely an identifier of an entity. Instead, the `EntityAccessor`
+directly provides methods to get, set, and check the existence
+of components, so that we do not need to call the world's 
+methods for this.
```
mlange-42, line 102: Maybe mention that access through queries is faster than access through the world?

docs/src/guide/queries_iteration.md
```diff
@@ -0,0 +1,292 @@
...
+        + String(pos[].x) + ", " + String(pos[].y) + ")"
+    )
+    if entity.has[Velocity]():
+        # Reset position and velocity
+        entity.set(Position(0, 0), Velocity(0, 0))
+```
+
+> [!Note] 
+> The `EntityAccessor` is a temporary object that is
+> created for each iteration. Therefore, should not be
```
mlange-42, line 118:
```suggestion
> created for each iteration. Therefore, it should not be
```

docs/src/guide/queries_iteration.md
```diff
@@ -0,0 +1,292 @@
...
+> stored in a container.
+
+> [!Note] 
+> Adding or removing entities or components to/from entities
+> would invalidate the iterator. Therefore, the world is 
+> locked during the iteration. This means that the forbidden
```
mlange-42, line 124:
```suggestion
> locked during the iteration. This means that these forbidden
```

docs/src/guide/queries_iteration.md
```diff
@@ -0,0 +1,292 @@
...
+We may want to apply a certain operation to all entities
+that have certain components. This can be achieved with 
+the {{< api World.apply apply >}} method. This method
+iterates over all entities conforming to a query and
+calls the provided function with the entities as arguments.
+The function must take a {{< api MutableEntityAccessor >}} 
+as its only argument.
+
+For example, if we want to apply a function that moves all entities
+with a `Position` and a `Velocity` component, we can do this as follows:
```
mlange-42, line 138: It remains unclear what the advantage of `apply` is, compared to the usual iteration.

docs/src/guide/queries_iteration.md
```diff
@@ -0,0 +1,292 @@
...
+```mojo {doctest="guide_simd_apply" hide=true}
+world = World[Position, Velocity]()
+_ = world.add_entities(Position(0, 0), Velocity(1, 0), count=10)
+```
+
+Before defining the vectorized operation, we need to gather some 
+information about the memory layout of our components.
+Specifically, we need to know the space in memory between
+two consecutive components of the same type. This is 
+needed if we want to load the same attribute of multiple components
```
mlange-42, line 206: Maybe be more explicit, like "the same component of multiple entities", or something like this? Had to think twice what might be meant here.

docs/src/guide/queries_iteration.md
```diff
@@ -0,0 +1,292 @@
...
+```mojo {doctest="guide_simd_apply"}
+alias stride = 2
+```
+
+Now we can define our vectorized move operation.
+It needs to take an integer parameter `simd_width`,
+which denotes how many entities will be processed at once.
+
+```mojo {doctest="guide_simd_apply"}
+fn move[simd_width: Int](entity: MutableEntityAccessor) capturing:
```
mlange-42, line 235: Wasn't there a simpler example in the API docs or so? I find this really hard to follow.

docs/src/guide/_index.md
samufi: Done.

docs/src/guide/queries_iteration.md
```diff
@@ -0,0 +1,292 @@
...
+
+    # Store the updated positions back to memory
+    pos_x_ptr.strided_store[width=simd_width](pos_x, stride)
+    pos_y_ptr.strided_store[width=simd_width](pos_y, stride)
+```
+
+What remains is to apply the move operation to all entities.
+For the vectorized version, the {{< api World.apply apply >}} method
+requires us to provide the `simd_width` parameter, which 
+denotes the maximal number of entities that can be processed
```
mlange-42, line 269: Maximum?

docs/src/guide/queries_iteration.md
```diff
@@ -0,0 +1,292 @@
...
+> itself. This is due to current limitations of Mojo and
+> will be changed as soon as possible.
+
+> [!Caution]
+> The world is locked during the iteration, and 
+> accessing variables outside a locally defined
+> function is an immature feature in Mojo. Do not
+> attempt to access the `world` from inside the operation.
+
+## Application of vectorized functions
```
mlange-42, line 169: As this is quite an advanced feature, I would (slightly) tend to move it to a separate chapter.

docs/src/guide/resources.md
```diff
@@ -0,0 +1,149 @@
++++
+type = "docs"
+title = "Resources"
+weight = 50
++++
+
+Not all data in a world is associated with 
+specific entities. Some data is shared across all entities
+(such as parameters), or is used to store global state
```
mlange-42, line 9: I would omit the `shared across all entities`, as it is not really clear what it means. Parameters could also be simple seen as "globals".

An typical illustrative example could be a global grid, or any spatial acceleration data structure, like a kd-tree.

docs/src/guide/queries_iteration.md
samufi: Done.

docs/src/guide/queries_iteration.md
samufi: Fixed.

docs/src/guide/resources.md
```diff
@@ -0,0 +1,149 @@
...
+
+```mojo {doctest="guide_resources"}
+# Add the `Time resource
+world.resources.add(Time(0.0))
+```
+The `resources` attribute also allows you to access and
+change resources via {{< api Resources.get get >}}, 
+{{< api Resources.get_ptr get_ptr >}} and 
+{{< api Resources.set set >}} methods resembling their
+[respective counterparts](../changing_entities) of `World`.
```
mlange-42, line 75: Maybe "component-related conterparts"? With the current wording, it is not clear what is meant.

docs/src/guide/queries_iteration.md
samufi: This has been removed. Please consider the new section on locks.

docs/src/guide/resources.md
```diff
@@ -0,0 +1,149 @@
...
+)
+
+# Set multiple resources
+world.resources.set(
+    Temperature(30.0),
+    Time(2.0) 
+)
+```
+
+As seen above with `SelectedEntities`, resources can also
```
mlange-42, line 123: Maybe "shown" or "demonstrated" instead of "seen"?
Maybe add "in contrast to components" or something like that?

docs/src/guide/systems_scheduler.md
```diff
@@ -0,0 +1,154 @@
++++
+type = "docs"
+title = "Systems and the scheduler"
+weight = 60
++++
+
+A key feature of entity-component systems is that 
+operations on the entities are organized in systems,
+which operate independently from one another and can
+be added or removed according to the respective needs.
```
mlange-42, line 10: "respective" is not the right word here in my (limited) understanding. Maybe "can be added or removed as required"?

docs/src/guide/systems_scheduler.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Systems
+
+Systems can be thought of as functions that take a 
+{{< api World >}} instance and perform operations on the 
+world's entities. However, to allow storing intermediate
+variables between multiple system calls, and to 
+support special initializing and finalizing operations
```
mlange-42, line 18:
```suggestion
support special initialization and finalization operations
```

docs/src/guide/systems_scheduler.md
```diff
@@ -0,0 +1,154 @@
...
+in the correct order. A `Scheduler` contains a {{< api World >}} instance
+and a list of systems. The systems are executed in the order they
+are added to the scheduler. The scheduler has 
+{{< api Scheduler.initialize initialize >}},
+{{< api Scheduler.update update >}}, and {{< api Scheduler.finalize finalize >}} 
+methods, which call the respective functions of all
+considered systems in the order they are added to the scheduler. 
+In addition, the scheduler has a {{< api Scheduler.run run >}}
+method, which initializes the systems, runs them a desired 
+number of times, and finalizes them.
```
mlange-42, line 66: Not related to the docs, but still...

Simply running for a given number of steps my not be sufficient. E.g., a simulation may be terminated when the simulated population goes extinct. A game may run indefinitely until the user quits, wins or looses.

In Ark (or rather, `ark-tools`), I do this via a resource which contains a bool that systems can set to signal that the run can terminate.

So I think you will need some way to allow systems to terminate a (potentially infinite) run.

docs/src/guide/queries_iteration.md
samufi: I understand the point, but I think this is clear from the text and I do not want to add further examples / think the examples at hand have their value.

docs/src/guide/queries_iteration.md
samufi: The API example was too far from the real world to be useful in a user guide. It showed how the function worked but not how this would be applied in useful examples.

samufi: I will see if I can make things more understandable, but this is an advanced feature requiring some knowledge about Mojo.

docs/src/guide/entities_components_world.md
```diff
@@ -0,0 +1,112 @@
...
```
mlange-42, line 18:
```suggestion
Larecs🌲 is an entity component system (ECS) framework for Mojo.
The ECS concept is composed of three principal elements:

- [Entities](#entities) represent game objects or simulation entities, like individuals in a population model.
- [Components](#components) are the data associated to entities, i.e. their properties or state variables.
- [Systems](../systems_scheduler) contain the game or simulation logic that manipulates entities and their components, using so-called queries.

In an ECS, each entity is "composed of" an arbitrary set of components that can be added and removed at run-time.
This modular design enables the development of highly flexible and reusable games or simulations.
By decoupling the logic (systems) from the data (entities and components),
ECS avoids convoluted inheritance hierarchies and eliminates hard-coded behavior.
Instead, entities can be composed of components to exhibit diverse behaviors without sacrificing modularity.

An ECS engine manages this architecture within a central storage structure known as the "[World](#the-world)".
The engine handles common tasks such as maintaining entity lists, spawning or deleting entities,
and scheduling logic operations, simplifying the development process for users.

Larecs🌲 provides a high-performance ECS framework,
empowering developers to create games and simulation models with exceptional flexibility and efficiency.
```

For attribution, it would be nice if you just press "accept" (in case you agree) instead of incorporating the changes manually.

docs/src/guide/queries_iteration.md
samufi: I will revise this section completely

docs/src/guide/queries_iteration.md
samufi: The example is now split into chunks to make it more understandable.

docs/src/guide/resources.md
samufi: Done.

docs/src/guide/resources.md
samufi: Done.

docs/src/guide/resources.md
samufi: Done.

docs/src/guide/systems_scheduler.md
samufi: Done.

docs/src/guide/systems_scheduler.md
```diff
@@ -0,0 +1,154 @@
...
+A key feature of entity-component systems is that 
+operations on the entities are organized in systems,
+which operate independently from one another and can
+be added or removed according to the respective needs.
+
+## Systems
+
+Systems can be thought of as functions that take a 
+{{< api World >}} instance and perform operations on the 
+world's entities. However, to allow storing intermediate
```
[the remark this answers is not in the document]
samufi: Done.

docs/src/guide/systems_scheduler.md
```diff
@@ -0,0 +1,154 @@
...
+variables between multiple system calls, and to 
+support special initializing and finalizing operations
+called at the beginning or the end of an ECS run, respectively,
+systems are expressed as structs implementing the 
+{{< api System >}} trait. This trait requires the
+systems to implement {{< api System.initialize initialize >}}, and
+{{< api System.finalize finalize >}} methods, called before or
```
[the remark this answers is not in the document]
samufi: Fixed.

docs/src/guide/systems_scheduler.md
```diff
@@ -0,0 +1,154 @@
...
+    fn finalize(mut self, mut world: World) raises:
+        # We do not need to do anything here
+        pass
+```
+
+## Scheduler
+
+The {{< api Scheduler >}} is responsible for executing the systems
+in the correct order. A `Scheduler` contains a {{< api World >}} instance
+and a list of systems. The systems are executed in the order they
```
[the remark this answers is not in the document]
samufi: Fixed.

docs/src/guide/queries_iteration.md
```diff
@@ -194,11 +194,11 @@ that have certain components. This can be achieved with
 the {{< api World.apply apply >}} method. This method
 iterates over all entities conforming to a query and
 calls the provided function with the entities as arguments.
-The function must take a {{< api MutableEntityAccessor >}} 
+The function must take a {{< api EntityAccessor MutableEntityAccessor >}}
```
mlange-42, line 197:
```suggestion
The function must take a mutable {{< api EntityAccessor EntityAccessor >}} 
```

Not sure, but there is no `MutableEntityAccessor` in the docs.

docs/src/guide/queries_iteration.md
```diff
@@ -194,11 +194,11 @@ that have certain components. This can be achieved with
 the {{< api World.apply apply >}} method. This method
 iterates over all entities conforming to a query and
 calls the provided function with the entities as arguments.
-The function must take a {{< api MutableEntityAccessor >}} 
+The function must take a {{< api EntityAccessor MutableEntityAccessor >}} 
 as its only argument. Applying a function to all entities
 can be more convenient and also faster than iterating over the entities
 manually, especially if the function is vectorized, as will be shown
```
mlange-42, line 200: No future tense... 😉

samufi: I don't understand. Is there future tense here?

docs/src/guide/queries_iteration.md
samufi: It is, but it is an alias for `EntityAccessor[True]`. Therefore, it does not have its own page.

docs/src/guide/queries_iteration.md
samufi: Ah, I see it.

samufi: Done.

mlange-42: As I commented on the other "this will be explained...". It *is* already explained, readers can follow the link right now, and don't need to wait until you explain it, some time in the future.

docs/src/guide/queries_iteration.md
mlange-42: So it is a "mutable `EntityAccessor`". I can't find `MutableEntityAccessor` anywhere in the docs.

samufi: Have a look here: http://localhost:1313/larecs/larecs/#aliases
Without this specific alias, the user might not know how to specify that it is mutable.

mlange-42: Ah, ok. Just seen that you also use the alias in the example. Well then, maybe link to the aliases instead of the struct?

mlange-42: Maybe I would even tend to use `EntityAccessor[True]` in the example, and remove the alias. Or would this require something like `EntityAccessor[True, _, _, _, _]`?

samufi: Fixed.

PR #99, opened 2025-04-08
Move resources instead of copying them

### LINKED ISSUE #100: adding and getting resources invokes copies
This is bad for performance.

For adding, a move should be sufficient. The copy is due to the usage of variadic arguments. Using specific overloads could help but are kind of ugly and limited.

For getting, this should not be the case! It is likely due to a bug in dict that could be worked around....

samufi:
Moving is much more performant than copying when considering large resources. 

Fixes #100.

Other things changed here:
- Add changelog
- Update readme
- Update Mojo and adjust Larecs accordingly

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
--
+- Move resources into the storage instead of copying them. This is much more performant if resources are large.
 
 ## v0.1.0 (2025-04-08)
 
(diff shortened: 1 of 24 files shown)


changelog.md
```diff
@@ -0,0 +1,9 @@
+# Changelog
+
+## Unreleased
```
mlange-42, line 3:
```suggestion
## [Unreleased](https://github.com/samufi/larecs/compare/v0.1.0...main)
```

mlange-42 (approved):

PR #103, opened 2025-06-02
Automatic resource identification

samufi:
Resources are automatically identified by their name, using the new builtin reflections utility. 
In addition, `Resources.get_ptr` has been removed, as this will not be necessary given the new `ref` keyword for storing references.


src/larecs/resource.mojo
```diff
@@ -1,56 +1,33 @@
 from collections import Dict
+from compile.reflection import get_type_name
```
mlange-42, line 2: The docs say "Returns the struct name of the given type parameter."  
So, is this safe for structs with the same name from different packages/modules?

samufi: Yes, the modules are separated with `::` afaik.

samufi: Just confirmed this again.

samufi: (though it uses `.` as separator, so it is exactly the convention we had earlier)

mlange-42 (approved):

Issue #45, opened 2025-01-19
Implement batched operations


samufi:
Some operations should potentially be executed for many entities at once to save time.
- add_entity
- remove_entity
- add (component)
- remove (component)
- replace (component)

I suggest implementing add_entity, as overload to the existing function, where a `count` argument would be provided. 

For the other methods I see two potential options.

### 1. OOP style

We could implement the operations as methods of a Batch object that could be returned by query. That way, the API stays slim users could do operations as follows:

```python
for entity in world.query[Position, Velocity]():
   ...

world.query[Position, Velocity]().add[Colour]()
for entity in world.query[Position, Velocity]().add[Colour]():
   ...
```

Of course, remove_entity would return an empty iterator. 

### 2. ECS style

We could also hand batches / filters over to add and remove etc. in the same way as an entity ID. 

```python
batch = world.get_batch[Position, Velocity]()

batch2 = world.add(batch, Colour("red"))
for entity in batch2:
   ...
```
We would need to be careful what happens to a batch that has been used. A suggestion would be to consume batch in functions that change the world and make it non-copyable.

Any thoughts @mlange-42?

mlange-42:
I would probably not move this functionality to queries, like in the "OOP style". I find is counter-intuitive.

Also, it is probably good to have an API as close to the normal operations as possible, so your `ECS style` example already looks quite good. But do you really need an additional type for the batches? Why not use queries here, too?

Another question is whether it is really useful to be able to assign in this operation (as in the 2nd example) or just allow parameters for the types (as in the 1st example). In most cases, users will iterate the created batch anyway to assign components to individual entities.

How about this? It is basically just "OOP style" flipped.

```mojo
for entity in world.Add[Rotation, Altitude](world.query[Position, Velocity]()):
    ...
```

What is the first line here? Just another usage example without a query loop?

```mojo
world.query[Position, Velocity]().add[Colour]()
for entity in world.query[Position, Velocity]().add[Colour]():
   ...
```

PR #113, opened 2025-08-07
Batch addition

moseschmiedel:
# Description
This implements batch addition of components for a given query. This is done by overloading the already existent `world.add` function with another version that takes a query instead of an entity (see `world.mojo` line 759ff).

# Default values for added components?
At the moment the function accepts default values for the added components. In #45 it is proposed to specify the added components only as parameters and leave component initialization to the user by iterating over all changed entities with the returned iterator. Should we do this?
As far as I can tell, the default setting of the component data impacts performance not significantly, as all entities need to be iterated anyway to add them to the new archetype and adjust the `EntityIndex` in the entity list.


src/larecs/world.mojo
```diff
@@ -756,6 +756,144 @@ struct World[*component_types: ComponentType](
...
+                new_archetype_index = self._get_archetype_index(
+                    component_ids, old_archetype[].get_node_index()
+                )
+
+                new_archetype = Pointer(
+                    to=self._archetypes.unsafe_get(new_archetype_index)
+                )
+
+                arch_start_idx = len(new_archetype[])
+                new_archetype[].reserve(arch_start_idx + len(old_archetype[]))
```
samufi, line 867: ensure that if a new allocation happens, the buffer size is increased by at least factor 2 so as to avoid frequent reallocations

samufi: Maybe it makes sense to build this constraint into the reserve function... Not sure but could be a good idea.

src/larecs/world.mojo
```diff
@@ -756,6 +756,144 @@ struct World[*component_types: ComponentType](
...
+
+                # save arch_start_idx for the iterator
```
samufi, line 869: I would start each inline coment with a capital letter unless it continues a sentence from the upper line

src/larecs/world.mojo
```diff
@@ -756,6 +756,144 @@ struct World[*component_types: ComponentType](
...
+                arch_start_idxs.append(arch_start_idx)
```
samufi, line 870: arch_start_idxs -> arch_start_idcs for better Engish spelling

src/larecs/world.mojo
```diff
@@ -756,6 +756,144 @@ struct World[*component_types: ComponentType](
...
+                    new_archetype[].unsafe_set(
+                        arch_start_idx + i,
+                        id,
+                        old_archetype[]._data[id],
+                        len(old_archetype[]),
+                    )
+
+                # add new components to the new archetype
+                @parameter
+                for comp_idx in range(add_components.__len__()):
```
samufi, line 885: I think you should be able to use `len()`

src/larecs/world.mojo
```diff
@@ -756,6 +756,144 @@ struct World[*component_types: ComponentType](
...
+
+                # save arch_start_idx for the iterator
+                arch_start_idxs.append(arch_start_idx)
+
+                # move component data from old archetype to new archetype
+                for i in range(old_archetype[]._component_count):
+                    id = old_archetype[]._ids[i]
+
+                    new_archetype[].unsafe_set(
+                        arch_start_idx + i,
```
samufi, line 877: Why do you add the index of the component to the start index of the entity?

samufi: In case this is an error, you could say `for id in old_archetype[]._component_count:`

moseschmiedel:
Found a bug, where entities on the old archetype are not getting removed, therefore converted this to draft again.

samufi (changes requested):

src/larecs/world.mojo
moseschmiedel: Yeah, this is mistake. This should only be `new_archetype[].unsafe_set(arch_start_idx, ...)`!

samufi (changes requested):
Great work, only small issues. Could you please also add a benchmark for batch add?

samufi (changes requested):
Could you add a benchmark of batch add?

moseschmiedel:
> Could you add a benchmark of batch add?

Sure.

src/larecs/archetype.mojo
```diff
@@ -432,20 +432,23 @@ struct Archetype[
...
+            new_capacity: The new minimal capacity of the archetype.
         """
         if new_capacity <= self._capacity:
             return
 
         for i in range(self._component_count):
             id = self._ids[i]
             old_size = index(self._item_sizes[id]) * self._capacity
-            new_size = index(self._item_sizes[id]) * new_capacity
+            new_size = index(self._item_sizes[id]) * new_capacity * 2
```
mlange-42, line 451: Does it really make sense to increase the capacity here again? Isn't the calculation already done on the caller's side?

BTW, looking at the resp. calculation in line 732:

```mojo
new_capacity = max(self._size + count, UInt(2) * self._capacity)
```

we get "odd" values if the intended capacity is more than twice the current one. In Ark, I always grow to the next power-of-two value.

src/larecs/query.mojo
```diff
@@ -392,22 +395,104 @@ struct _ArchetypeIterator[
...
+            )
         )
+        self._index += 1
+
+    fn __len__(self) -> Int:
+        """
+        Returns the number of archetypes remaining in the iterator.
+
+        Note that this requires iterating over all archetypes
+        and may be a complex operation.
```
mlange-42, line 493: Leftover from C&P iterator over entities?

src/larecs/world.mojo
```diff
@@ -756,6 +812,177 @@ struct World[*component_types: ComponentType](
...
+        # If query could match archetypes that already have at least one of the components, raise an error
+        if not query.without_mask[].contains(BitMask(component_ids)):
+            raise Error(
+                "Query could match entities that already have at least one of"
+                " the components to add. Use `Query.without[Component, ...]()`"
+                " to exclude those components."
+            )
+
+        arch_start_idcs = List[UInt, True](len(self._archetypes))
+        changed_archetype_idcs = List[Int, True](len(self._archetypes))
```
mlange-42, line 894: It might be overly pessimistic to reserve memory for the total number of archetypes in the world. Maybe set a (generous) upper limit and risk growing the list later?

moseschmiedel: Yes, I was thinking that the total number of archetypes will never be that big. Even `255 * 8 byte ~ 2kb` which is nothing, keeping in mind that current L1 caches have hundreds of kilobyte (e.g. see https://www.amd.com/de/support/downloads/drivers.html/processors/ryzen/ryzen-5000-series/amd-ryzen-5-5600.html).
What would your proposed initial capacity be?

src/larecs/archetype.mojo
moseschmiedel: I think I don't completely understand what you're suggesting. Do you think we should do the calculation inside `reserve` (like it is implemented in this PR) or increase the capacity only, so that the `new_capacity` fits?

mlange-42: In this PR, you set the capacity to twice the requested capacity (i.e. twice the one given as the argument), which does not really make sense IMO. The variant in line 732 looks better in that it doubles the current capacity if necessary. However, it does so only once, not until the new capacity matches or exceeds the requested one. So if the current capacity is 128 and you request 277, you get 277 instead of 512. And it does so outside of `reserve`, so the result gets doubled again (at least I think so, without looking at the code again).

My suggestion would be to do such calculations only in `reserve`, never outside of it, and to double the current (not the requested) capacity until sufficient, like this (or with some log2+pow calculation or bit-shift trick):

```mojo
if new_capacity <= self._capacity:
    return

cap = self._capacity
while cap < new_capacity:
    cap *= 2

# do the actual work...
```

src/larecs/world.mojo
```diff
@@ -756,6 +812,177 @@ struct World[*component_types: ComponentType](
...
+        #    must be excluded in the query. Therefore, we can apply the transformation to each matching archetype
+        #    individually without checking for edge cases where multiple archetypes get merged into one.
+        #    This also enables potential parallelization optimizations.
+
+        self._assert_unlocked()
+
+        alias component_ids = Self.component_manager.get_id_arr[*Ts]()
+
+        # If query could match archetypes that already have at least one of the components, raise an error
+        if not query.without_mask[].contains(BitMask(component_ids)):
```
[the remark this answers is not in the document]
moseschmiedel: I implemented a naive check where all matching existing archetypes are iterated and checked if they have conflicting components. This is quite a bit more costly then the BitMask check (`O(n)` instead of `O(1)`) but it does not complain as often as the other implementation ...and it is still faster then individual `world.add` calls according to the benchmark.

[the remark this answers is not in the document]
moseschmiedel: What are your thoughts on this implementation?

src/larecs/world.mojo
```diff
@@ -756,6 +867,192 @@ struct World[*component_types: ComponentType](
...
+
+        for archetype in self._get_archetype_iterator(
+            query.mask, query.without_mask
+        ):
+            if query.matches(archetype[].get_mask()):
+                if not archetype[].get_mask().contains(
+                    BitMask(component_ids)
+                ) and archetype[].get_mask().contains_any(
+                    BitMask(component_ids)
+                ):
```
mlange-42, line 952: 1. Isn't the outer condition (`query.matches(...)`) already covered by the iterator?
1. If the second part of the inner condition gets evaluated, you construct the mask twice.
1. I do not really understand the first part og the inner condition. Why is it ok that the archetype already contains all the components that are to be added?

A general comment/question: do you (particularly @samufi) assume that all raised errors can be caught and recovered? My approach in Ark is to panic on "misuse", which makes it easier to do checks like this "on the fly" and not needing to check everything before making changes.

src/larecs/world.mojo
```diff
@@ -756,6 +867,192 @@ struct World[*component_types: ComponentType](
...
+        alias component_ids = Self.component_manager.get_id_arr[*Ts]()
+
+        # If query could match archetypes that already have at least one of the components, raise an error
+
+        for archetype in self._get_archetype_iterator(
+            query.mask, query.without_mask
+        ):
+            if query.matches(archetype[].get_mask()):
+                if not archetype[].get_mask().contains(
+                    BitMask(component_ids)
```
samufi, line 949: Is there really a need for a cast in each for loop iteration?

moseschmiedel: No, but AFAIK this is really easy for the compiler to optimize and therefore there should be no need for manual optimization here.

src/larecs/world.mojo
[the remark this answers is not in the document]
samufi: @mlange-42 I do not understand how the stricter version could lead to faulty reasoning, as it requires a more explicit statement. Could you elaborate / give an example?
I think having the more expensive check is fine, but I would wrap it with a check of the simpler condition. Then people who want to be explicit can do that and save time that way.

src/larecs/world.mojo
samufi: Interesting. How reliable are such things?

src/larecs/world.mojo
mlange-42: Ok, this makes sense somehow.

But maybe it would be useful to add a user guide section on the "error philosophy"? Particulary stating that catching errors should not be a "usual thing", but that Larecs raises errors only on problems that are to be considered either misuse or mistakes in the model logoc.

moseschmiedel: > 1. Isn't the outer condition (query.matches(...)) already covered by the iterator?
Yes, this a left-over from a previous idea I had...😅

> 2. If the second part of the inner condition gets evaluated, you construct the mask twice.
I think it wouldn't because the compiler could optimize that. See my comment on the message from @samufi 

> 3. I do not really understand the first part of the inner condition. Why is it ok that the archetype already contains all the components that are to be added?
It is ok, because then it is one of the target archetypes. It wouldn't lead to multiple source archetypes mapping on to one target archetype. Consider following example
```
Archetypes: A, AB, AC, ABC
add(query[A], B, C)
```
In this example archetype `ABC` is no problem because it is the target of `A + BC`. But `AB` and  `AC` are problems, because they would also get transformed to `ABC`, which leads to following mapping:
```
A -> ABC
AB -> ABC
AC -> ABC
```
We don't want this, because then we have dependencies between the loop iterations, which would prevent parallization later on.

mlange-42: So adding B+C to ABC is a no-op (or rather, replaces the components)?

IMO it is not a good idea to allow this, as it is counter-intuitive and potentially error-prone. Also, allowing to replace all but not allowing to replace some and add others seems inconsistent to me.

src/larecs/world.mojo
```diff
@@ -756,6 +867,192 @@ struct World[*component_types: ComponentType](
...
+                ) and archetype[].get_mask().contains_any(
+                    BitMask(component_ids)
+                ):
+                    raise Error(
+                        "Query could match entities that already have at least"
```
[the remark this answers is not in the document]
moseschmiedel: No, I changed that!

src/larecs/world.mojo
```diff
@@ -756,6 +867,192 @@ struct World[*component_types: ComponentType](
...
+                        " one of the components to add. Use"
+                        " `Query.without[Component, ...]()` to exclude those"
+                        " components."
+                    )
+
+        alias _2kb = (1024 * 2) // sizeof[UInt]()
```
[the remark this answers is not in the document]
moseschmiedel: Is the name of the alias not descriptive enough for you? What sort of information do you want to have here as a comment?

src/larecs/world.mojo
```diff
@@ -756,6 +867,189 @@ struct World[*component_types: ComponentType](
...
+        #    must be excluded in the query. Therefore, we can apply the transformation to each matching archetype
+        #    individually without checking for edge cases where multiple archetypes get merged into one.
+        #    This also enables potential parallelization optimizations.
+
+        self._assert_unlocked()
+
+        alias component_ids = Self.component_manager.get_id_arr[*Ts]()
+
+        # If query could match archetypes that already have at least one of the components, raise an error
+        if not query.without_mask.or_else(BitMask()).contains(
```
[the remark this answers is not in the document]
samufi: Note, we are not working with Optional here but StaticOptional. Use an `@parameter if`, then this should work.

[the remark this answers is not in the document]
moseschmiedel: No, then the second condition cannot be part of the if statement anymore, because it depends on a dynamic value.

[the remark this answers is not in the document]
samufi: Couldn't you solve this via an intermediate boolean value, which is constant True if not has_without_mask and else-wise query.without_mask[].contains(...)? The extra value may be optimized away on the constant branch...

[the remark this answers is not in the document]
moseschmiedel: Do you mean like that?
```suggestion
        foo = False
        
        @parameter
        if has_without_mask:
            foo = True

        if foo or not query.without_mask[].contains(
```

This triggers the same bug mentioned earlier.

samufi (approved):

mlange-42 (approved):

PR #118, opened 2025-09-24
Migration to Mojo 0.25.6

moseschmiedel:
See https://docs.modular.com/mojo/changelog#v0256-2025-09-22


samufi:
I would argue that everything that holds a somewhat constant state should be implicitly copyable and everything else not. So I would see PitPool and _ArchetypeIterator on the explicit side. But we can see if this makes ergonomics bad...

Component is not a struct I believe...

moseschmiedel:
> I would argue that everything that holds a somewhat constant state should be implicitly copyable and everything else not. So I would see PitPool and _ArchetypeIterator on the explicit side. But we can see if this makes ergonomics bad...

If we allow `ImplicitCopyable` on `BitPool` we can let the compiler synthesize the `__copyinit__` for `LockMask` for us. Otherwise we need to write it ourselves. Does that change your opinion on that?

samufi:
> > Component is not a struct I believe...
> 
> You are right, it is an alias to `Copyable & Movable` at the moment. The question is, if we require `ImplicitlyCopyable` here aswell?

I would require as little as possible here, unless it enforces correctness somehow. So if there was a trait for trivial types I would use that. But ImplicitlyCopyable does not suffice, so I would leave it away.

samufi:
> > I would argue that everything that holds a somewhat constant state should be implicitly copyable and everything else not. So I would see PitPool and _ArchetypeIterator on the explicit side. But we can see if this makes ergonomics bad...
> 
> If we allow `ImplicitCopyable` on `BitPool` we can let the compiler synthesize the `__copyinit__` for `LockMask` for us. Otherwise we need to write it ourselves. Does that change your opinion on that?

I actually think it is the other way round. LockMask should not be implicitly copyable, as this is the one instance that contains all the locks. The name somehow suggests that it is simply some kind of boolean array, which is why I thought it should be trivital. Maybe refactoring here could be good. Maybe to `Locker` or `Lock`?

changelog.md
```diff
@@ -3,11 +3,42 @@
 ## [Unreleased](https://github.com/samufi/larecs/compare/v0.4.0...main)
 
 ### Breaking changes
+- Update the utilized Mojo version to 25.6 and adjust the code accordingly.
+- Revisit what structs are only `Copyable` and which can be also `ImplicitlyCopyable`
```
samufi: Revisit *which* ...

changelog.md
```diff
@@ -3,11 +3,42 @@
...
+  - UnsafeBox
+  - World
+
+  #### ImplicitlyCopyable
+  - BitMask
+  - Component
+  - Entity
+  - EntityIndex
+  - LockedContext
+  - LockMask -> LockManager
```
samufi: This is not implicitly copyable, is it?

changelog.md
```diff
@@ -3,11 +3,42 @@
...
+
 - Rename `_ArchetypeIterator` to `_ArchetypeByMaskIterator`
+- Rename `LockMask` to `LockManager`
+- Remove all `hint_trivial_type` and `run_destructors` parameters from containers that leaked them from their underlying
+  List attribute
 
 ### Other changes
 - Implement batch component addition as overload of `world.add`
 - Implement batch component removal as overload of `world.remove`
+- Remove `unsafe_take`
```
samufi: from the _utils module

src/larecs/bitmask.mojo
```diff
@@ -191,12 +193,12 @@ struct BitMask(Copyable, EqualityComparable, KeyElement, Movable, Stringable):
...
     fn contains(self, other: Self) -> Bool:
         """Reports if the other mask is a subset of this mask."""
-        return ((self._bytes & other._bytes) == other._bytes).reduce_and()
+        return (self._bytes & other._bytes) == other._bytes
 
     @always_inline
     fn contains_any(self, other: Self) -> Bool:
         """Reports if any bit of the other mask is in this mask."""
-        return ((self._bytes & other._bytes) != 0).reduce_or()
+        return (self._bytes & other._bytes) != 0
```
samufi, line 201: Isn't this doing an implicit reduce_and?

src/larecs/query.mojo
```diff
@@ -102,13 +113,11 @@ struct Query[
             Error: If the lock cannot be acquired (more than 256 locks exist).
         """
         iterator = self._world[]._get_entity_iterator(
-            self._mask, self._without_mask
+            self._mask, self._without_mask.copy()
```
samufi: Maybe we should make _get_entity_iterator take the without_mask not as owned. I think there is actually no real benefit of it being copied. (Also, it makes the code ugly)

src/larecs/query.mojo
```diff
@@ -324,7 +343,7 @@ struct _ArchetypeByMaskIterator[
         """
         query_info = Self.QueryInfo(
             mask=self._mask,
-            without_mask=self._without_mask,
+            without_mask=self._without_mask.copy(),
```
samufi: Maybe also here let us not use an owned mask but rather copy the provided value in the constructor.

src/larecs/resource.mojo
```diff
@@ -57,57 +48,57 @@ struct Resources(ExplicitlyCopyable, Movable, Sized):
...
 
+        conflicting_ids = List[StringSlice[StaticConstantOrigin]]()
+
+        @parameter
+        for idx in range(resources.__len__()):
+            alias id = get_type_name[Ts[idx]]()
+            if id in self._storage:
+                conflicting_ids.append(id)
+
+        if len(conflicting_ids) > 0:
```
samufi: Could also be done with an implicit bool cast.

src/larecs/resource.mojo
```diff
@@ -57,57 +48,57 @@ struct Resources(ExplicitlyCopyable, Movable, Sized):
...
+            id: The ID of the resource to add. It has to be not used already.
             resource: The resource to add.
-
-        Raises:
-            Error: If the resource already exists.
         """
-        if id in self._storage:
-            raise Error("Resource already exists.")
-        self._storage[id] = UnsafeBox(resource[])
+        self._storage[id] = UnsafeBox(resource[].copy())
```
samufi: This always makes a copy when the resource is added. Do we find a way to move resources into Reseources (at least if the ^ operator is used)?

samufi: Actually, I never see this overload used. Let us just delete it then.

src/larecs/resource.mojo
```diff
@@ -122,40 +113,49 @@ struct Resources(ExplicitlyCopyable, Movable, Sized):
...
+                get_type_name[Ts[idx]](),
+                resource^,
             )
-        __disable_del resources
+
+        resources^.consume_elements[take_resource]()
 
     @always_inline
     fn _set[
         T: Copyable & Movable, add_if_not_found: Bool
```
samufi: Is it possible now to make T infer only, so that we do not have to explicitly write add_if_not_found=add_if_not_found later? (There used to be a bug, but maybe it has been fixed?)

src/larecs/world.mojo
```diff
@@ -466,8 +427,8 @@ struct World[*component_types: ComponentType](
             @parameter
             for i in range(size):
                 archetype[].get_component[
-                    T = Ts[i.value], assert_has_component=False
-                ](index_in_archetype) = components[i]
+                    T = Ts[i], assert_has_component=False
+                ](index_in_archetype) = components[i].copy()
```
samufi, line 431: Maybe a dumn question, but assigning a copied value does not copy twice, does it? (like, evelaute the rhs, then copy the value to the lhs)

src/larecs/world.mojo
```diff
@@ -1579,7 +1538,7 @@ struct World[*component_types: ComponentType](
             for archetype in _ArchetypeByMaskIterator(
                 Pointer(to=self._archetypes),
                 query.mask,
-                query.without_mask,
+                query.without_mask.copy(),
```
samufi: Maybe we can also here take the without_mask not as owned...

src/larecs/world.mojo
```diff
@@ -1697,9 +1656,9 @@ struct World[*component_types: ComponentType](
         has_without_mask: Bool = False, has_start_indices: Bool = False
     ](
         mut self,
-        owned mask: BitMask,
-        owned without_mask: StaticOptional[BitMask, has_without_mask],
-        owned start_indices: _EntityIterator[
+        var mask: BitMask,
```
samufi: Let us rethink which of the elements really need to be takes as owned. Otherwise we are spending a long time writing .copy() - at least for the elements that are not implicitly copyable...

samufi (changes requested):
Good job! Only minor things we may want to think about...

changelog.md
moseschmiedel: No, it is not.

src/larecs/bitmask.mojo
moseschmiedel: No, first `self._bytes & other._bytes` results in a new `SIMD[DType.uint8, Self.total_bytes]` where only the bits are 1 that occur in both `BitMask`s. Comparing this vector to 0 is equal to asking if any bit is 1 which in turn is only the case if the bit was 1 in `self._bytes` and `other._bytes`.

src/larecs/query.mojo
moseschmiedel: I think it is good to leave this decision to the user of `_get_entity_iterator`. In the end we always need an owned `without_mask` as `_ArchetypeByMaskIterator` wants one. With the current API the user of `_get_entity_iterator` can decide if he wants to copy some value they already have into the iterator or if they even want to move the value. When we decide to not take this argument as owned, we always have to do the copy internally which could be optimized away in some cases previously.

src/larecs/query.mojo
moseschmiedel: Same argument as with `_get_entity_iterator`.

src/larecs/bitmask.mojo
samufi: OK, understood. == becomes reduce_and and != becomes reduce_or.

src/larecs/query.mojo
samufi: To protocol our off-line discussion: in this particular case, we gain ergonomics by making this not owned but do not loose control, as the type is trivial and needs to be copied completely even in move operation.

src/larecs/query.mojo
moseschmiedel: Removed the owned requirement here as well to increase ergonomics.

src/larecs/world.mojo
moseschmiedel: I assume this to behave the same as if I would construct a new Component on the rhs (`archetype[]... = Component(42)`). In this case there is no separate copy, the newly constructed Component is directly initialized in the memory pointed to by the ref on the lhs. Extrapolating from this the explicit copy should also initialize itself directly in the memory pointed to by the ref. See https://docs.modular.com/mojo/manual/lifecycle/life#copy-constructor.

src/larecs/world.mojo
moseschmiedel: Going from the previous discussion about ownership of `without_mask` I would also change `mask` to be unowned. `start_indices` should stay owned IMO because it is a `List` which is not cheap to copy.

benchmark/plots/src/aos.mojo
```diff
@@ -210,7 +210,7 @@ fn run_benchmarks(config: BenchConfig) raises -> List[BenchResult]:
             result = benchmark[compExp](rounds, entities)
             results.append(result)
 
-    return results
+    return results^
```
samufi: Instead of moving the result out, we should use the out argument convention and directly fill the output slot.

src/larecs/resource.mojo
```diff
@@ -57,57 +48,42 @@ struct Resources(ExplicitlyCopyable, Movable, Sized):
...
         Raises:
-            Error: If the resource already exists.
+            Error: If some resource already exists.
         """
 
-        @parameter
-        for i in range(resources.__len__()):
-            self._add(get_type_name[Ts[i]](), unsafe_take(resources[i]))
-        __disable_del resources
+        conflicting_ids = List[StringSlice[StaticConstantOrigin]]()
```
samufi: We could initialize the list with a minimal capacity (0?), as it is unlikely that there are conflicting IDs - and if there are some, an extra allocation is not a problem. Alternatively we could also start by first checking if anything is duplicate and only then allocate a list to return the IDs where this is an issue. I don't think this is a real problem, but please consider it.

samufi (changes requested):
Very minor changes.

src/larecs/resource.mojo
moseschmiedel: Yes, I think the current version is more or less already doing what you are suggesting. If I forced the minimal capacity to 0 the first "real" allocation (besides allocating space for the pointer and size of the list, which are on the stack) happens when the first conflicting ID is found. Note that the problem here is not looking for duplicates in `*Ts` but to look for already registered resources in `self._storage`. So to do the check before allocating the list would mean that we need to first iterate through `resources` until we find a conflict and then allocate the list and then continue iterating. This is nearly equivalent to allocating the list with capacity=0, iterating through `resources` and appending all conflicts.

benchmark/world_benchmark.mojo
```diff
@@ -472,6 +472,56 @@ fn prevent_inlining_add_remove_1_comp() raises:
...
+    pos = Position(1.0, 2.0)
+    comp = FlexibleComponent[1](1, 42.0)
+
+    @always_inline
+    @parameter
+    fn bench_fn() capturing raises:
+        world = SmallWorld()
+
+        # create 1_000_000 entities that initially do not have FlexibleComponent[1]
+        _ = world.add_entities(pos, count=1_000_000)
```
samufi: I think the benchmakr below covers this already and we do not need two benchmarks for this. I'd suggest removing it.

benchmark/world_benchmark.mojo
```diff
@@ -537,11 +587,14 @@ fn benchmark_batch_add_1_comp_1_000_000(
         _ = world.add(
             world.query[Position]().without[FlexibleComponent[1]](), comp
         )
+        _ = world.remove[FlexibleComponent[1]](
```
samufi: If I am not mistaken, this does as much add_entities as add and remove. I'd suggest doing the add and remove 1000 times and working with 1000 entities only.

benchmark/world_benchmark.mojo
```diff
@@ -573,6 +626,75 @@ fn benchmark_batch_add_5_comp_1_000_000(
             comp4,
             comp5,
         )
```
samufi: As above: I think it is sufficient to keep the 1000 x 1000 benchmark

AGENTS.md
```diff
@@ -0,0 +1,31 @@
...
+- Run all tests: `pixi run mojo test -I src/ test/`
+- Run single test: `pixi run mojo test -I src/ test/<filename>.mojo`
+- Format code: `pixi run mojo format src test benchmark`
+- Generate docs: `pixi run mojo doc -o docs/src/larecs.json src/larecs`
+
+## Code Style
+- Use snake_case for functions/variables, PascalCase for types
+- Use `fn` for static functions, `def` for dynamic functions
+- Prefer `var` for mutable, immutable by default
+- Use `inout` parameters for mutation, not return modified values
```
samufi: Use `mut` not `inout`

AGENTS.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Build/Test Commands
+- Run all tests: `pixi run mojo test -I src/ test/`
+- Run single test: `pixi run mojo test -I src/ test/<filename>.mojo`
+- Format code: `pixi run mojo format src test benchmark`
+- Generate docs: `pixi run mojo doc -o docs/src/larecs.json src/larecs`
+
+## Code Style
+- Use snake_case for functions/variables, PascalCase for types
+- Use `fn` for static functions, `def` for dynamic functions
```
samufi: For now, always use `fn`. We may use `def` in main and raising test functions.

AGENTS.md
```diff
@@ -0,0 +1,31 @@
...
+- Prefer `var` for mutable, immutable by default
+- Use `inout` parameters for mutation, not return modified values
+- Include comprehensive type hints with Mojo's progressive typing
+- Use `@parameter` for compile-time constants, `@always_inline` for critical paths
+- Leverage SIMD types `SIMD[type, width]` for vectorization
+- Apply traits: `Copyable`, `Movable`, `Stringable` appropriately
+- Use manual memory management with `UnsafePointer` when needed
+- Include docstrings for public APIs
```
samufi: Every function should have a docstring including a description of all parameters, raises and returns.

AGENTS.md
```diff
@@ -0,0 +1,31 @@
...
+- Reference Mojo docs for LLMs: https://docs.modular.com/llms-mojo.txt
+
+## Error Handling & Safety
+- Follow borrow checker principles for memory safety
+- Prefer stack allocation and RAII patterns
+- Use `debug_warn()` utility for debug messages
```
samufi, line 47: maybe also use `debug_assert` for critical checks that should not impact performance

samufi (approved):

PR #119, opened 2025-09-26
Batch replace

moseschmiedel:


moseschmiedel:
Something is causing the tests to fail in the pipeline. I cannot reproduce these failures locally. @samufi do you have a guess where the error comes from? Maybe somewhere in the `ArchetypeByListIterator` code, but technically `__has_next__` should prevent out-of-bounds access to the List attributes...?

Edit: should be fixed now

benchmark/world_benchmark.mojo
```diff
@@ -463,7 +463,7 @@ fn benchmark_add_remove_1_comp_1_000_000(
     bencher.iter[bench_fn]()
 
 
-fn prevent_inlining_add_remove_1_comp() raises:
+fn prevent_inlining_add_remove_1_comp_1_000_000() raises:
```
samufi: If this is only to prevent inlining, we do not need to add the number of calls (1_000_000) to the name. Same in the examples below.

benchmark/world_benchmark.mojo
```diff
@@ -496,7 +496,21 @@ fn benchmark_add_remove_1_comp_batch_1_000_000(
...
 
 
-fn benchmark_add_remove_1_comp_1_000_batch_1000(
+fn prevent_inlining_add_remove_1_comp_batch_1_000_000() raises:
+    pos = Position(1.0, 2.0)
+    comp = FlexibleComponent[1](1, 42.0)
+    world = SmallWorld()
+
+    # create 1_000_000 entities that initially do not have FlexibleComponent[1]
+    _ = world.add_entities(pos, count=1_000_000)
```
samufi: This can be done with an arbitrary small batch, as this is just to add some call to the function. I would only use a small batch of 1.

benchmark/world_benchmark.mojo
```diff
@@ -522,6 +536,24 @@ fn benchmark_add_remove_1_comp_1_000_batch_1000(
     bencher.iter[bench_fn]()
 
 
+fn prevent_inlining_add_remove_1_comp_1_000_batch_1_000() raises:
```
samufi: I think all the functions in here are also called in the previous function, right? Then we don't need this twice.

benchmark/world_benchmark.mojo
```diff
@@ -699,6 +751,51 @@ fn benchmark_add_remove_5_comp_1_000_batch_1000(
...
+    comp2 = FlexibleComponent[2](1.0, 42.0)
+    comp3 = FlexibleComponent[3](1.0, 42.0)
+    comp4 = FlexibleComponent[4](1.0, 42.0)
+    comp5 = FlexibleComponent[5](1.0, 42.0)
+    world = SmallWorld()
+
+    # create 1_000 entities that initially do not have FlexibleComponent[1...5]
+    _ = world.add_entities(pos, count=1_000)
+    # then 1_000 x add components and remove them afterwards
+    for _ in range(1000):
```
samufi: I don't see a point for for loops in the functions to prevent inlining...

docs/src/guide/changing_entities.md
```diff
@@ -173,10 +173,51 @@ for entity in entities:
...
+    world.query[Position, Velocity]()
+)
+```
+
+The query must ensure that all matching entities have the components you want to remove,
+otherwise an error will be raised.
+
+#### Batch replacing components
+
+You can replace components on multiple entities that match a query using the
```
samufi: components *of*

docs/src/guide/changing_entities.md
```diff
@@ -173,10 +173,51 @@ for entity in entities:
     world.add(entity, Velocity(1.0, 0.5))  # Individual operations
 ```
 
-> [!Note]
-> Currently, only batch adding of components is supported. 
-> Batch removal and replacement operations are planned for future releases.
-> See the [roadmap](../../../README.md#next-steps) for more information.
+#### Batch removing components
```
samufi: "Batch-removing" (and earlier "Batch-adding")

docs/src/guide/changing_entities.md
```diff
@@ -173,10 +173,51 @@ for entity in entities:
...
+You can replace components on multiple entities that match a query using the
+{{< api World.replace replace >}} method in combination with {{< api Replacer.by by >}}:
+
+```mojo {doctest="guide_change_entities"}
+# Add 10 entities with Position components
+_ = world.add_entities(Position(0, 0), count=10)
+
+# Replace Position with Velocity for all entities that have Position
+world.replace[Position]().by(
+    world.query[Position](),
```
samufi: Maybe we should encourage using the other overload (with an explicit `query` keyword argument), because I think this is more readable.

src/larecs/bitmask.mojo
```diff
@@ -168,6 +158,46 @@ struct BitMask(
         else:
             self._bytes[index(idx)] &= ~(1 << offset)
 
+    @always_inline
+    fn set(self, *comps: Self.IndexType, value: Bool) -> Self:
+        """Modifies the [BitMask] to unset the components given as arguments."""
```
samufi: Why "unset" here and below?

src/larecs/bitmask.mojo
```diff
@@ -227,7 +262,64 @@ struct BitMask(
         This operation is highly optimized using SIMD instructions for fast parallel
         bitwise operations across all 256 bits simultaneously.
         """
-        return BitMask(bytes=self._bytes | other._bytes)
+        copy = self.copy()
```
samufi: use the output slot to avoid another copy on return

src/larecs/bitmask.mojo
```diff
@@ -227,7 +262,64 @@ struct BitMask(
...
+            other: The other BitMask to AND with this mask.
+
+        Returns:
+            A new BitMask containing the bitwise AND of both masks.
+
+        **Performance Note:**
+        This operation is highly optimized using SIMD instructions for fast parallel
+        bitwise operations across all 256 bits simultaneously.
+        """
+        copy = self.copy()
```
samufi: As above, use the out slot

src/larecs/bitmask.mojo
```diff
@@ -210,6 +240,11 @@ struct BitMask(
         """Returns the indices of the bits that are set."""
         result = _BitMaskIndexIter(self._bytes)
 
+    @always_inline
+    fn __invert__(self) -> BitMask:
```
samufi: Use `Self` as return type.

src/larecs/bitmask.mojo
```diff
@@ -210,6 +240,11 @@ struct BitMask(
         """Returns the indices of the bits that are set."""
         result = _BitMaskIndexIter(self._bytes)
 
+    @always_inline
```
samufi: We should think about a default order for functions. Maybe the dunders first and then the other ones? What do you think?

src/larecs/world.mojo
```diff
@@ -101,6 +101,113 @@ struct Replacer[
...
         )
 
+    fn by[
+        *AddTs: ComponentType,
+        has_without_mask: Bool = False,
+    ](
+        self,
+        query: QueryInfo[has_without_mask=has_without_mask],
+        *components: *AddTs,
+        out iterator: World[*component_types].Iterator[
```
samufi: A less verbose alternative for the return type:

```
__type_of(self._by(
            components,
            query=query,
        ))
```

I think this is not much less readable, but it is easier to maintain if a change needs to be made to self._by.

src/larecs/world.mojo
```diff
@@ -101,6 +101,113 @@ struct Replacer[
...
+            Error: when called with components that can't be added because they are already present.
+            Error: when called with components that can't be removed because they are not present.
+            Error: when called on a locked world. Do not use during [.World.query] iteration.
+        """
+        return self._by(
+            components,
+            query=query,
+        )
+
+    fn by[
```
samufi, line 155: same as above

src/larecs/world.mojo
```diff
@@ -101,6 +101,113 @@ struct Replacer[
...
+        )
+
+    fn _by[
+        *AddTs: ComponentType,
+        has_without_mask: Bool = False,
+    ](
+        self,
+        components: VariadicPack[_, _, ComponentType, *AddTs],
+        query: QueryInfo[has_without_mask=has_without_mask],
+        out iterator: World[*component_types].Iterator[
```
samufi: even here, we could use the return type of _batch_remove_and_add.

src/larecs/world.mojo
```diff
@@ -1276,99 +1157,87 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
-
-        var archetype_index: Int = -1
-        compare_mask = old_archetype[].get_mask()
-
-        alias add_error_msg = "Entity already has one of the components to add."
-        alias remove_error_msg = "Entity does not have one of the components to remove."
+        old_archetype_mask = old_archetype[].get_mask()
 
         @parameter
         if rem_size:
```
samufi, line 1241: can we say `if remove_some and rem_size:` for better readability?

src/larecs/world.mojo
```diff
@@ -1276,99 +1157,87 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
-                )
-            else:
-                archetype_index = self._get_archetype_index(
-                    remove_ids[], start_node_index
-                )
-                # No need for Pointer revalidation due to previous memory reservation!
+            if remove_some:
+                if not old_archetype_mask.contains(BitMask(remove_ids[])):
+                    raise Error(
+                        "Entity does not have one of the components to remove."
```
samufi:
```suggestion
                        "Entity does not have all the components to remove."
```

src/larecs/world.mojo
```diff
@@ -1276,99 +1157,87 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
 
-        archetype = Pointer(to=self._archetypes.unsafe_get(archetype_index))
-        index_in_archetype = archetype[].add(entity)
+            @parameter
+            if remove_some:
+                compare_mask = compare_mask.set(remove_ids[], False)
+            if compare_mask.contains(BitMask(add_ids)):
+                raise Error("Entity already has one of the components to add.")
+
+        alias ComponentIdsType = InlineArray[Self.Id, add_size + rem_size]
```
samufi, line 1255: What happens if remove_some is false and rem_size > 0? Maybe we should introduce an alias for remove_some * rem_size?

samufi: Or we add a constraint, since the above should never happen and the case would be a sign dor something going wrong.

samufi: In this case, we only need to check one of the two variables (also in the if statements further up)

src/larecs/world.mojo
```diff
@@ -1276,99 +1157,87 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+        var component_ids: ComponentIdsType
 
         @parameter
-        if add_size:
-            if not archetype[].get_mask().contains(compare_mask):
-                raise Error(add_error_msg)
+        if add_size and rem_size:
+            component_ids = ComponentIdsType(uninitialized=True)
```
samufi: maybe introduce an always_inlined concatenate function for this task in the _utils module? I think this would be clearer.

samufi: And this may be able to run completely at compiletime if you do the copy with  @parameter for loops.

src/larecs/world.mojo
```diff
@@ -1377,19 +1246,309 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+            remove_ids:     The IDs of the components to remove.
+
+        Returns:
+            An iterator over the modified entities.
+
+        Raises:
+            Error: when called with nothing to do (i.e. no components to add or remove).
+            Error: when called on a locked world. Do not use during [.World.query] iteration.
+        """
+        alias add_size = add_components.__len__()
```
samufi, line 1399: Please go through the comments to the previous function and have a look if they apply here as well.

src/larecs/world.mojo
```diff
@@ -1377,19 +1246,309 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+            if has_without_mask:
+                strict_check_needed = not query.without_mask[].contains(
+                    BitMask(add_ids)
+                )
+            else:
+                strict_check_needed = True
+
+            if strict_check_needed:
+                for archetype in self._get_archetype_iterator(
+                    query.mask, query.without_mask.copy()
```
samufi:
```suggestion
                    query.mask, query.without_mask
```

src/larecs/world.mojo
```diff
@@ -1377,19 +1246,309 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+
+                    @parameter
+                    if remove_some:
+                        archetype_mask = archetype_mask.set(remove_ids[], False)
+
+                    if archetype[] and archetype_mask.contains_any(
+                        BitMask(add_ids)
+                    ):
+                        raise Error(
+                            "Query matches entities that already have at least"
```
samufi: "some of the ..."

src/larecs/world.mojo
```diff
@@ -1377,19 +1246,309 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+                            " `Query.without[Component, ...]()` to exclude"
+                            " those components."
+                        )
+
+        @parameter
+        if rem_size:
+            # If query could match archetypes that don't have all of the components, raise an error
+            if not query.mask.contains(BitMask(remove_ids[])):
+                raise Error(
+                    "Query matches entities that don't have all of the"
```
samufi:
```suggestion
                    "Query matches entities that do not have all of the"
```

src/larecs/world.mojo
```diff
@@ -1377,19 +1246,309 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+                    " components to remove. Use `Query(Component, ...)` to"
+                    " include those components."
+                )
+
+            @parameter
+            if has_without_mask:
+                if query.without_mask[].contains_any(BitMask(remove_ids[])):
+                    raise Error(
+                        "Query excludes entities that have a component which"
+                        " should be removed in the without mask. Remove all"
```
samufi: Without masks are an internal implementation detail. Hide this in the error message and say what the problem is from the user's perspective

src/larecs/world.mojo
```diff
@@ -1377,19 +1246,309 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+                component_ids.unsafe_ptr(),
+                remove_ids[].unsafe_ptr(),
+                rem_size,
+            )
+            memcpy(
+                component_ids.unsafe_ptr() + rem_size * size_of[Self.Id](),
+                add_ids.unsafe_ptr(),
+                add_size,
+            )
+        elif Bool(add_size) and not rem_size:
```
samufi, line 1463: Do we need the explicit cast to bool?

src/larecs/world.mojo
```diff
@@ -1377,19 +1246,309 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+        remove_some: Bool = False,
+        has_without_mask: Bool = False,
+    ](
+        mut self,
+        query: QueryInfo[has_without_mask=has_without_mask],
+        *add_components: *Ts,
+        remove_ids: StaticOptional[
+            InlineArray[Self.Id, rem_size], remove_some
+        ] = None,
+        out iterator: Self.Iterator[
```
samufi, line 1326: Could also be simplified with __type_of

src/larecs/world.mojo
```diff
@@ -1377,19 +1246,309 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+                Pointer(to=self._locks),
+                Self.ArchetypeIterator[
+                    __origin_of(self._archetypes),
+                    arch_iter_variant_idx=_ArchetypeByListIteratorIdx,
+                ](
+                    Self.ArchetypeByListIterator[__origin_of(self._archetypes)](
+                        Pointer(to=self._archetypes), List[Int]()
+                    ),
+                ),
+                StaticOptional(List[UInt]()),
```
samufi: No explicit construction of StaticOptional necessary

src/larecs/world.mojo
```diff
@@ -1377,19 +1246,309 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+                component_ids.unsafe_ptr() + rem_size * size_of[Self.Id](),
+                add_ids.unsafe_ptr(),
+                add_size,
+            )
+        elif Bool(add_size) and not rem_size:
+            component_ids = rebind[ComponentIdsType](add_ids)
+        elif not add_size and Bool(rem_size):
+            component_ids = rebind[ComponentIdsType](remove_ids[])
+        else:
+            return Self.Iterator[
```
samufi: This may be a good place to use initializer lists... (and also in similar situations above and below.)

samufi (changes requested):
I have not finished reviewing the actual logic of the replace function, but there are some other things to consider already. Nothing serious, though.

src/larecs/bitmask.mojo
moseschmiedel: C&P mistake...

src/larecs/bitmask.mojo
moseschmiedel: I agree. Sadly the stdlib seems not to have an opinion in this matter. I think dunders first is a good default. And if we want we can always change it later.

src/larecs/world.mojo
moseschmiedel: Yes, as `not` casts the rhs to `Bool` and `and` is only implemented when lhs and rhs are both `Bool` or both `UInt`.

src/larecs/world.mojo
moseschmiedel: Sadly this does not work here, because somehow the compiler cannot correctly resolve the overload and thinks that `__type_of(self._batch_remove_and_add(query, add_components, remove_ids))` is a recursive reference. Even if I explicitly cast the second argument (`self._batch_remove_and_add(query, VariadicPack(add_components), remove_ids))` it fails. So for now we have to leave that as is.

docs/src/guide/changing_entities.md
```diff
@@ -173,10 +173,51 @@ for entity in entities:
     world.add(entity, Velocity(1.0, 0.5))  # Individual operations
```
samufi: I think the counterexample (how to do it without batch) is not needed and makes the section longer than required. I would remove this.

Instead of adding an extra section for each operation (add, remove, replace), just give an example for each and try to emphasize the common pattern - and, where necessary, the differences. The goal is that the reader understands the idea behind this stuff rather than learning the commands by heart.

src/larecs/_utils.mojo
```diff
@@ -81,3 +82,22 @@ fn next_pow2[dtype: DType](var value: Scalar[dtype]) -> Scalar[dtype]:
         value |= value >> (2**i)
 
     return value + 1
+
+
+@always_inline
+fn concatenate_inline_arrays[
```
samufi, line 88: I think this function should have a test.

src/larecs/bitmask.mojo
```diff
@@ -113,19 +113,104 @@ struct BitMask(
...
+        Performs element-wise bitwise OR operation between this mask and another mask,
+        creating a new mask where a bit is set if it's set in either operand.
+
+        Args:
+            other: The other BitMask to OR with this mask.
+
+        Returns:
+            A new BitMask containing the bitwise OR of both masks.
+
+        **Performance Note:**
```
samufi: Are the performance notes AI-generated? I do not think they are necessary and puff up the code base. I suggest removing them.

src/larecs/bitmask.mojo
```diff
@@ -168,6 +253,48 @@ struct BitMask(
         else:
             self._bytes[index(idx)] &= ~(1 << offset)
 
+    @always_inline
```
samufi, line 369: This module was written before the automated doc gen was in place. Therefore, the doc strings do not conform to the larecs standard. Could you adjust the comments to consider all arguments, parameters, and returns? Maybe you could use an AI promt that does this for the entire file? =D

src/larecs/bitmask.mojo
```diff
@@ -178,7 +305,7 @@ struct BitMask(
     @always_inline
```
samufi: If this is simply an alias for `~` we could also remove it. Or what do you think?

src/larecs/bitmask.mojo
```diff
@@ -113,19 +113,104 @@ struct BitMask(
...
+        for i in range(len(self._bytes) * 8):
+            if self.get(i):
+                result += "1"
+            else:
+                result += "0"
+        result += "]"
+        return result
+
+    @always_inline
+    fn __repr__(self) -> String:
```
samufi, line 283: IF you figure out how the writer thing workd, it would be awesome to be able to implement all the things necessary to put this into a `print(mask)` function. (but this does not have priority)

src/larecs/world.mojo
```diff
@@ -28,13 +28,42 @@ from .query import (
...
 from .lock import LockManager, LockedContext
 from .resource import Resources
+from ._utils import concatenate_inline_arrays
+
+
+struct WorldErrors:
+    alias non_existent_entity_msg = "The considered entity does not exist anymore."
+    alias world_is_locked_msg = "Attempt to modify a locked world."
+    alias missing_components_for_removal_entity_msg = "Entity does not have all the components to remove."
+    alias missing_components_for_removal_query_msg = "Query matches entities that do not have all the components to remove. Use `Query(Component, ...)` to include those components."
```
samufi: Is the syntax `Query(Component)` correct here? I think the components are parameters, right?

src/larecs/world.mojo
```diff
@@ -28,13 +28,42 @@ from .query import (
 )
 from .lock import LockManager, LockedContext
 from .resource import Resources
+from ._utils import concatenate_inline_arrays
+
+
+struct WorldErrors:
+    alias non_existent_entity_msg = "The considered entity does not exist anymore."
```
samufi, line 35: Is it necessay (or beneficial) to have the messages and errors as separate aliases? Otherwise, I find it more compact to store them together right away.

src/larecs/world.mojo
```diff
@@ -28,13 +28,42 @@ from .query import (
...
+        Self.duplicate_components_for_addition_query_msg
+    )
 
 
 @fieldwise_init
 struct Replacer[
     world_origin: MutableOrigin,
     size: Int,
     *component_types: ComponentType,
+    remove_ids: InlineArray[World[*component_types].Id, size],
```
samufi, line 63: Nice one! =)

src/larecs/world.mojo
```diff
@@ -95,10 +123,137 @@ struct Replacer[
...
+        *AddTs: ComponentType,
+        has_without_mask: Bool = False,
+    ](
+        self,
+        query: QueryInfo[has_without_mask=has_without_mask],
+        *components: *AddTs,
+        out iterator: __type_of(self._by(components, query=query)),
+    ) raises:
+        """
+        Removes and adds the components to a multiple [..entity.Entity] specified by a [..query.Query].
```
samufi: "to multiple entities"

src/larecs/world.mojo
```diff
@@ -112,6 +267,20 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
 
     alias Id = BitMask.IndexType
     alias component_manager = ComponentManager[*component_types]()
+    alias _rem_size[*Ts: ComponentType] = VariadicPack[
```
samufi: re _rem_size and _remove_ids: make rem and remove consistent. I suggest sticking with the full word

samufi: Actually, is this all specific to removal of components? This is basically a mechanism to map some variables. Maybe use a name closer to what it does rather than to where it is used.

src/larecs/world.mojo
```diff
@@ -1100,275 +1141,164 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
-                # 2. If an archetype with the new component combination does not exist yet,
-                #    create new archetype B = A - component_ids and move entities and component data from A to B.
-                new_archetype_idx = self._get_archetype_index(
-                    component_ids, old_archetype[].get_node_index()
-                )
+    @always_inline
+    fn _remove_and_add[
+        *Ts: ComponentType,
+        rem_size: Int = 0,
+        remove_some: Bool = False,
```
samufi: could remove_some be replaced by `rem_size > 0`?

src/larecs/bitmask.mojo
samufi: The same mistake is in the set function doc strings below. Search for the word "unset" and check if it really is supposed to be there.

src/larecs/bitmask.mojo
```diff
@@ -168,6 +253,48 @@ struct BitMask(
         else:
             self._bytes[index(idx)] &= ~(1 << offset)
 
+    @always_inline
+    fn set(self, *comps: Self.IndexType, value: Bool) -> Self:
+        """Modifies the [BitMask] to set the components given as arguments to the provided value.
```
samufi, line 371: Why does `set` have a return type? And why is self not mutable if the method modifies the BitMask?

src/larecs/world.mojo
```diff
@@ -1100,275 +1141,164 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
-            if add_size:
-                start_node_index = self._archetype_map.get_node_index(
-                    remove_ids[], start_node_index
-                )
-                if not compare_mask.contains(
-                    self._archetype_map.get_node_mask(start_node_index)
-                ):
-                    raise Error(remove_error_msg)
+            if remove_some:
+                compare_mask = compare_mask.set(runtime_remove_ids[], False)
```
samufi: set should be an in-place operation, right?

src/larecs/world.mojo
```diff
@@ -1100,275 +1141,164 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+        elif not add_size and Bool(rem_size):
+            component_ids = rebind[ComponentIdsType](runtime_remove_ids[])
         else:
-            if not compare_mask.contains(archetype[].get_mask()):
-                raise Error(remove_error_msg)
+            return
 
+        index_in_old_archetype = entity_index.index
+        new_archetype_idx = self._get_archetype_index(
+            component_ids, old_archetype[].get_node_index()
```
samufi, line 1273: The pointer old_archetype could have become invaldi, since the archetype list could have been resized when looking for the new archetype. (see the batch version of this function)

src/larecs/world.mojo
```diff
@@ -1377,21 +1307,295 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+                #    and insert new component data for moved entities.
+                # 2. If an archetype with the new component combination does not exist yet,
+                #    create new archetype B = A.different_by(component_ids) and move entities and component data from A to B.
+                new_archetype_idx = self._get_archetype_index(
+                    component_ids, old_archetype[].get_node_index()
+                )
+
+                # We need to update the pointer to the old archetype, because the `self._archetypes` list may have been
+                # resized during the call to `_get_archetype_index`.
+                old_archetype_idx = self._archetype_map[
```
samufi, line 1507: You trust that this is optimized out of the loop by the compiler?

src/larecs/world.mojo
```diff
@@ -1377,21 +1307,295 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+
+        # Search for the archetype that matches the query mask
+        with self._locked():
+            for old_archetype in self._get_archetype_iterator(
+                query.mask, query.without_mask
+            ):
+                # Two cases per matching archetype A:
+                # 1. If an archetype B with the new component combination exists, move entities from A to B
+                #    and insert new component data for moved entities.
+                # 2. If an archetype with the new component combination does not exist yet,
```
samufi, line 1499: I somehow did not spot a case distinction in the code. Could you help me with a line number? Also, note that the second case applies also if the new archetype has size 0. This is relevant if components are added and removed back and forth - which would become almost a free operation if this optimization is implemented.

benchmark/world_benchmark.mojo
```diff
@@ -721,6 +691,198 @@ fn benchmark_replace_1_comp_1_000_000(
...
+            FlexibleComponent[0](1.0, 2.0),
+            FlexibleComponent[1](3.0, 4.0),
+            FlexibleComponent[2](5.0, 6.0),
+            FlexibleComponent[3](7.0, 8.0),
+            FlexibleComponent[4](9.0, 10.0),
+            count=1_000,
+        )
+
+        for _ in range(500):
+            _ = world.replace[
```
samufi, line 849: Replacing components back and forth will make heavy use of the optimization that data do not need to be copied if the new archetype is empty. Think about if this is really waht you want to benchmark and if not, prevent this from happening (add a new entitity with the same components as the new archetype; remove this component later)

samufi (changes requested):
Thanks, this fixes many of the issue I raised earlier. I found some new things to consider - and one potential bug, if I am not mistaken =).

src/larecs/world.mojo
moseschmiedel: I stored them separately so that the strings can be used in the tests. See `world_test.mojo` line 252 for reference.

src/larecs/world.mojo
moseschmiedel: It shouldn't because the `self._archetypes` list could possibly have been resized at every run of the loop. So this precaution has to be taken at every iteration to ensure valid pointers.

src/larecs/world.mojo
moseschmiedel: The case distinction is hidden in the call to `self._get_archetype_index()` (Line 1501). There a new archetype is created when the combination of `component_ids` and `old_archetype[].get_node_index()` not yet exist.

src/larecs/world.mojo
moseschmiedel: This is already safe here because of line 1230 (archetype list is ensured to have enough space for at least one more entity).

src/larecs/world.mojo
moseschmiedel: Concerning your second point, this would need a change of `self._get_archetype_index()` so that it doesn't auto-create archetypes or don't use it here. I think my preference would be to change the behaviour of `self._get_archetype_index()` so that it does only one thing. What do you think @samufi?

docs/src/guide/changing_entities.md
```diff
@@ -168,17 +171,47 @@ This is significantly more efficient than adding components to entities one by o
...
+# You can also replace multiple components with multiple other components
+world.replace[Position, Velocity]().by(
+    world.query[Position, Velocity](),
+    Direction(0.0, 5.0),
+    Acceleration(0.2, 0.4)
+)
+```
+
+> [!Important]
+> The query must ensure that all matching entities have the components you want to remove,
```
samufi:
```suggestion
> The query must ensure that all matching entities have the components you want to remove;
```

docs/src/guide/changing_entities.md
```diff
@@ -168,17 +171,47 @@ This is significantly more efficient than adding components to entities one by o
...
+world.remove[Position, Velocity](
+    world.query[Position, Velocity]()
+)
 ```
 
-> [!Note]
-> Currently, only batch adding of components is supported. 
-> Batch removal and replacement operations are planned for future releases.
-> See the [roadmap](../../../README.md#next-steps) for more information.
+For batch replace operations you also have to use the {{< api Replacer.by by >}} helper method to specify which
```
samufi:
```suggestion
For batch replace operations, you also need to use the {{< api Replacer.by by >}} helper method to specify which 
```

src/larecs/world.mojo
samufi: I am afraid I don't get it. Sure, a new archetype is created only if no such archetype exists yet. But the optimization is in not copying the data but simply moving them over. And I do not see a move of the archetype but only a copy of the data. So where does the move happen?

samufi: Regarding the second point, why would `self._get_archetype_index()` make a difference? I would simply say that if the new archetype is empty, then I move the data containers (for the components) over. Otherwise, I go through the individual entities and copy their entries. What am I missing?

src/larecs/world.mojo
```diff
@@ -112,6 +262,27 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
 
     alias Id = BitMask.IndexType
     alias component_manager = ComponentManager[*component_types]()
+    alias _components_size[*Ts: ComponentType] = VariadicPack[
```
samufi:
```suggestion
    alias _component_count[*Ts: ComponentType] = VariadicPack[
```

samufi: "Size" could also refer to the size in memory. I think "component count" captures that this counts how many components there are in the list.
Alternatively, we could also define an internal `_len[*Ts: AnyType]` or `component_len[*Ts: ComponentType]` to show that this is not a constant for the world but depends on the variadics.
Maybe adjust `_components_is_empty` to `_is_empty` then.

samufi (commented):
Mostly trivial stuff only. However, I did not see / understand the move optimiation when the new archetype's size is 0. If this is clarified, we can go ahead.

pixi.toml
```diff
@@ -1,5 +1,8 @@
-[project]
-authors = ["Samuel M. Fischer <samufi@web.de>", "Martin Lange <martin_lange_@gmx.net>"]
+[workspace]
+authors = [
+    "Samuel M. Fischer <samufi@web.de>",
+    "Martin Lange <martin_lange_@gmx.net>",
```
samufi, line 4: You should add your name here, @moseschmiedel.

moseschmiedel:
@samufi I added the optimization

src/larecs/world.mojo
```diff
@@ -1377,21 +1296,293 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+                index_in_new_archetype,
+                add_ids[i],
                 UnsafePointer(to=add_components[i]).bitcast[UInt8](),
             )
 
         swapped = old_archetype[].remove(index_in_old_archetype)
         if swapped:
-            var swapEntity = old_archetype[].get_entity(idx.index)
-            self._entities[swapEntity.get_id()].index = idx.index
+            var swapEntity = old_archetype[].get_entity(entity_index.index)
```
samufi: use snake_case: `swap_entity`

src/larecs/world.mojo
```diff
@@ -1377,21 +1296,293 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+        elif Bool(add_size) and not rem_size:
+            component_ids = rebind[ComponentIdsType](add_ids)
+        elif not add_size and Bool(rem_size):
+            component_ids = rebind[ComponentIdsType](runtime_remove_ids[])
+        else:
+            return {
+                Pointer(to=self._locks),
+                Self.ArchetypeIterator[
+                    __origin_of(self._archetypes),
+                    arch_iter_variant_idx=_ArchetypeByListIteratorIdx,
```
samufi, line 1472: May this be better solved with a proper enum?

src/larecs/world.mojo
```diff
@@ -1377,21 +1296,293 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+                    to=self._archetypes.unsafe_get(index(old_archetype_idx))
+                )
+
+                new_archetype = Pointer(
+                    to=self._archetypes.unsafe_get(new_archetype_idx)
+                )
+
+                # If the new archetype is empty, we can optimize component movement by just pointing to the component data in the old archetype without actually moving any data. We just need to make sure to properly initialize the new archetype's metadata and to reserve space for new components that get added.
+                if len(new_archetype[]) == 0:
+                    new_archetype[].shallow_copy_from(
```
samufi: Can this be encapsulated in Archetype? So that `shallow_copy_from` takes `old_archetype[]` directly?

src/larecs/archetype.mojo
```diff
@@ -397,6 +397,19 @@ struct Archetype[
         """
         return Bool(self._size)
 
+    fn reinit_components(mut self, ids: SIMD[Self.dType, Self.max_size]):
```
samufi: I think this is an unsafe operation, because it does not free the old memory. I think this method should be private or specifically marked as unsafe.

src/larecs/archetype.mojo
```diff
@@ -397,6 +397,19 @@ struct Archetype[
...
+        """
+        Reinitializes the data pointed to by `self._data` of the archetype without changing the component layout.
+        """
```
samufi, line 413: docs for the ids are missing

src/larecs/archetype.mojo
```diff
@@ -574,6 +572,39 @@ struct Archetype[
             index(self._item_sizes[id]) * count,
         )
 
+    @always_inline
+    fn shallow_copy_from(
```
samufi: Could we also make this take another archetype? (`take_data_from` and then apply `reinit_components`) Or do we get in trouble due to argument exclusivity?

I think there should typically never be two archetypes managing the same pointer. Therefore, this should always be labelled as a move or take operation.

samufi (changes requested):
Thanks for this. I wonder if some of the operations could be encapsulated a bit better. 

I noted that a move from one archetype to another remains an O(#entities) operation unless we really change the old archetype. This is someting to think about in the future, to make add and remove etc. an O(1) operation in benign cases.

src/larecs/archetype.mojo
moseschmiedel: This doesn't work due to the ownership semantics of Mojo (`old_archetype[]` and `new_archetype[]` both point to the same underlying list of archetypes in `World`).

I renamed it so that it is labelled as a take operation. I also added a note that the components in the old archetype need to be reinitialized.

src/larecs/world.mojo
moseschmiedel: See my answer to your question below

src/larecs/world.mojo
moseschmiedel: What do you mean by `a proper enum`?

samufi: I mean distinguishing the achetype iterator variants via a dedicated Type not an index, along with a type to index mapping function.

samufi: The idea is that users do not need to know about the correct indices.

src/larecs/archetype.mojo
samufi: OK, I can't say I like it, but for now let us leave it that way. Later we can think about whether an implicit cast to a data extractor could solve the problem.

samufi (changes requested):
I think we can go on with this. I would only like you to think about the enum thing first.

src/larecs/world.mojo
samufi: Scratch that. I have had a second look and I think it is fine as is. When updating, we will be able to use a built-in type for StaticVariant, I believe.

samufi (approved):
Ready to go.

PR #120, opened 2025-09-30
Query refactor

moseschmiedel:
Removed `QueryInfo` and use `filter.MaskFilter` instead. Only rely on matching logic implemented in `MaskFilter` to avoid duplication of that logic.


docs/src/guide/changing_entities.md
```diff
@@ -207,8 +207,8 @@ _ = world.add_entities(Position(0, 0), count=10)
 
 # Replace Position with Velocity for all entities that have Position
```
mlange-42, line 208:
```suggestion
# Replace Position by Velocity for all entities that have Position
```

For consistency with the API.

docs/src/guide/changing_entities.md
```diff
@@ -207,8 +207,8 @@ _ = world.add_entities(Position(0, 0), count=10)
...
 world.replace[Position]().by(
+    world.query[Position](),
     Velocity(2.0, 2.0)
-    query=world.query[Position](),
 )
```
mlange-42, line 212: Not sure whether this is possible, but it would probably read more fluently if the query was an argument of `replace` rather then `by`. "Replace position [for] query by velocity"

docs/src/guide/changing_entities.md
mlange-42, line 218: Maybe use different numbers of components in replace and by? Otherwise, it might cause the impression that the same number of entities must be removed and added.

mlange-42, line 223: This is stated in the introduction to batch operations already. But maybe the repetition and additional emphasis is ok.

src/larecs/bitmask.mojo
```diff
@@ -71,6 +71,13 @@ struct BitMask(
         """Initializes the mask with the given bytes."""
         self._bytes = bytes
 
+    @always_inline
+    fn __init__(out self, *bits: Self.IndexType):
+        """Initializes the mask with the bits at the given indices set to True.
+        """
+        self = Self(bits)
+
+    @implicit
```
mlange-42, line 80: I guess this allows to simply pass component ID varargs to functions/methods that take a mask?

src/larecs/bitmask.mojo
```diff
@@ -296,17 +238,19 @@ struct BitMask(
...
+        copy = self.copy()
+        copy.flip_mut(bit)
+        return copy
+
+    @always_inline
+    fn flip_mut(mut self, bit: Self.IndexType):
         """Flips the state of bit at the given index."""
         var idx: Self.IndexType = bit >> 3  # equivalent to bit // 8
         var offset: Self.IndexType = bit & 7  # equivalent to bit - (8 * idx)
         self._bytes[index(idx)] ^= 1 << offset
```
mlange-42, line 252: Maybe follow the doc pattern of the methods above, stating return/inplace explicitly?

src/larecs/world.mojo
```diff
@@ -123,26 +95,35 @@ struct Replacer[
...
+        out iterator: World[*component_types].Iterator[
+            __origin_of(self._world[]._archetypes),
+            __origin_of(self._world[]._locks),
+            arch_iter_variant_idx=_ArchetypeByListIteratorIdx,
+            has_start_indices=True,
+        ],
     ) raises:
         """
-        Removes and adds the components to a multiple [..entity.Entity] specified by a [..query.Query].
+        Removes and adds the components to a multiple [..entity.Entity] specified by a [..filter.MaskFilter].
```
mlange-42, line 119:
```suggestion
        Removes and adds the components to multiple [..entity.Entity] specified by a [..filter.MaskFilter].
```

src/larecs/world.mojo
```diff
@@ -152,28 +133,33 @@ struct Replacer[
...
```
mlange-42, line 154:
```suggestion
        Removes and adds the components to multiple [..entity.Entity] specified by a [..filter.MaskFilter].
```

src/larecs/world.mojo
```diff
@@ -824,17 +762,16 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
         ```
 
         Args:
-            query: The query to determine which entities to remove. Note, you can
-                   either use [..query.Query] or [..query.QueryInfo].
+            filter: The filter to determine which entities to remove. Note, you can
+                   either use [..filter.MaskFilter] or [..query.Query] because [..filter.MaskFilter]
```
mlange-42, line 766: Will `MaskFilter` be exposed as public API / included in the API docs? Why would that be useful? In case not, maybe simply state that a query can/should be used?

EDIT: As `MaskFilter` is currently not re-exported in `larecs/__init__.mojo`, it will not show up in the API docs.

src/larecs/world.mojo
```diff
@@ -1040,12 +980,12 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
         Parameters:
-            has_without_mask: Whether the query has a without mask.
+            has_exclude: Whether the filter has an exclude mask.
             Ts: The types of the components to add.
 
         Args:
-            query: The query specifying which entities to modify. The query must explicitly exclude existing entities
-                that already have some of the components to add.
+            filter: The [..filter.MaskFilter] specifying which entities to modify. The filter must explicitly exclude
+                existing entities that already have some of the components to add.
```
mlange-42, line 988: Is the "must explicitly exclude" still true?

Also, maybe mention the option to use a query here.

src/larecs/world.mojo
```diff
@@ -1074,30 +1014,25 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+            __origin_of(self._archetypes),
+            __origin_of(self._locks),
+            arch_iter_variant_idx=_ArchetypeByListIteratorIdx,
+            has_start_indices=True,
+        ],
     ) raises:
         """
-        Removes components from multiple entities at once, specified by a [..query.Query].
-        The provided query must ensure that matching entities have all of the components that should get removed.
+        Removes components from multiple entities at once, specified by a [..filter.MaskFilter].
```
mlange-42, line 1034: Also mention query here?

src/larecs/world.mojo
```diff
@@ -1249,33 +1175,45 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
         old_archetype_mask = old_archetype[].get_mask()
 
         @parameter
-        if remove_some:
-            if not old_archetype_mask.contains(BitMask(runtime_remove_ids[])):
-                raise WorldErrors.missing_components_for_removal_entity
+        if rem_size:
+
+            @parameter
+            if remove_some:
```
mlange-42, line 1181: Shouldn't this also be `@parameter`? Aren't `rem_size` and `remove_some` redundant information?

src/larecs/world.mojo
```diff
@@ -1249,33 +1175,45 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+                if not old_archetype_mask.contains(BitMask(remove_ids[])):
+                    raise Error(
+                        "Entity does not have one of the components to remove."
```
mlange-42, line 1184: This could be misunderstood in terms of that only one is required. Maybe better "Entity is missing one of the components to remove."?

src/larecs/world.mojo
```diff
@@ -1343,46 +1281,44 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
         ],
     ) raises:
         """
-        Adds and removes components to a multiple [..entity.Entity] specified by a [..query.QueryInfo].
+        Adds and removes components to a multiple [..entity.Entity] specified by a [..filter.MaskFilter].
```
mlange-42, line 1284:
```suggestion
        Adds and removes components to multiple [..entity.Entity] specified by a [..filter.MaskFilter].
```

mlange-42: Mention query for use as `MaskFilter` again?

src/larecs/world.mojo
```diff
@@ -1391,18 +1327,18 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
```
mlange-42, line 1330:
```suggestion
        Adds and removes components to multiple [..entity.Entity] specified by a [..filter.MaskFilter].
```

src/larecs/world.mojo
mlange-42, line 1505: No skipping of removed components here?

src/larecs/world.mojo
```diff
@@ -1636,7 +1586,7 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
                 (see [vectorize doc](https://docs.modular.com/mojo/stdlib/algorithm/functional/vectorize)).
 
         Args:
-            query: The query to determine which entities to apply the operation to.
+            filter: The [..filter.MaskFilter] to determine which entities to apply the operation to.
```
mlange-42, line 1589: Mention query here too?

src/larecs/world.mojo
```diff
@@ -1728,7 +1678,7 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
```
mlange-42, line 1681: Query?

src/larecs/world.mojo
```diff
@@ -1842,26 +1790,39 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
         Returns:
             A [..query.Query] for all entities with the given components.
         """
+
+        iterator = Self.Query(Pointer(to=self), self.filter[*Ts]())
+
+    @always_inline
+    fn filter[
+        *Ts: ComponentType
+    ](mut self, out mask_filter: MaskFilter[has_exclude=False]):
```
mlange-42, line 1799: If this is the way for users to create `MaskFilter`s, the type should probably be re-exported in `larecs/__init__.mojo` and included in the API reference.

src/larecs/bitmask.mojo
samufi: This is necessary as an overload if the caller has already received the varargs and needs to pass them down. Mojo does not have proper value unpacking as of yet.

src/larecs/bitmask.mojo
```diff
@@ -255,29 +147,70 @@ struct BitMask(
 
     @always_inline
     fn set(self, *comps: Self.IndexType, value: Bool) -> Self:
```
samufi, line 149: I find it awkward to have `set` not mutate self but rather return a copy. Is there a specific reason for that?

src/larecs/bitmask.mojo
```diff
@@ -336,3 +280,138 @@ struct BitMask(
...
+        """Performs in-place bitwise OR with another BitMask.
+
+        This method modifies the current BitMask by performing a bitwise OR operation
+        with another BitMask. Each bit in the resulting BitMask is set if it is set
+        in either the current BitMask or the provided BitMask.
+
+        Args:
+            other: The BitMask to perform the bitwise OR operation with.
+
+        **Performance Note:**
```
samufi, line 321: This note is unnecessary (here as well as in all other places), as it does not provide any useful information.

src/larecs/world.mojo
```diff
@@ -1249,33 +1175,45 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+                    )
 
         @parameter
         if add_size:
             compare_mask = old_archetype_mask
 
             @parameter
             if remove_some:
-                compare_mask = compare_mask.set(runtime_remove_ids[], False)
+                compare_mask = compare_mask.set(remove_ids[], False)
```
samufi, line 1193: You could use `set[False](...)` here, couldn't you?

src/larecs/world.mojo
```diff
@@ -1441,55 +1368,75 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
+                remove_ids[].unsafe_ptr(),
+                rem_size,
+            )
+            memcpy(
+                component_ids.unsafe_ptr() + rem_size * size_of[Self.Id](),
+                add_ids.unsafe_ptr(),
+                add_size,
             )
-            component_ids = rebind[ComponentIdsType](concatenated)
         elif Bool(add_size) and not rem_size:
```
samufi, line 1429: no explicit casting required. Remove.

src/larecs/world.mojo
```diff
@@ -1441,55 +1368,75 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
             component_ids = rebind[ComponentIdsType](add_ids)
         elif not add_size and Bool(rem_size):
-            component_ids = rebind[ComponentIdsType](runtime_remove_ids[])
+            component_ids = rebind[ComponentIdsType](remove_ids[])
         else:
-            return {
+            return Self.Iterator[
```
samufi, line 1434: Maybe use initializer lists for shorter and more readable code.

src/larecs/world.mojo
```diff
@@ -1583,7 +1528,12 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
                 old_archetype[].clear()
 
         # Return iterator to iterate over the changed entities.
-        iterator = {
+        iterator = Self.Iterator[
```
samufi, line 1531: Use initializer list here.

src/larecs/world.mojo
samufi: I think this should be removed. If users can always use query, the API is much simpler and nicer, and the overhead is minimal. So, I think we should remove this function.

docs/src/guide/changing_entities.md
samufi: We would need to store the filter in the replacer object. This is doable in principle, but has a small overhead. @moseschmiedel what do you think?
The currently implemented version is consistent with the version for single entities, so we would need to change both (and also for the single entity we'd have a small runtime overhead, which might be more significant there). 
Maybe a benchmark could help deciding?

src/larecs/bitmask.mojo
```diff
@@ -296,17 +238,19 @@ struct BitMask(
...
     @always_inline
-    fn flip(mut self, bit: Self.IndexType):
+    fn flip(self, bit: Self.IndexType) -> Self:
+        """Flips the state of bit at the given index."""
+        copy = self.copy()
+        copy.flip_mut(bit)
+        return copy
+
+    @always_inline
+    fn flip_mut(mut self, bit: Self.IndexType):
```
samufi, line 248: I think, `flip` should do the mutable operation and we should think of another name for the operation that returns a copy. For example `get_flipped` or `copy_flipped` or `flipped_copy`.

benchmark/world_benchmark.mojo
```diff
@@ -463,17 +463,13 @@ fn benchmark_add_remove_1_comp_1_000_000(
     bencher.iter[bench_fn]()
 
 
-fn prevent_inlining_add_remove() raises:
+fn prevent_inlining_add_remove_1_comp_1_000_000() raises:
```
samufi, line 466: Not that it matters, but the inlining prevention works regardless of how many entities the actual benchmark considers. So I'd keep the old name.

benchmark/world_benchmark.mojo
```diff
@@ -500,6 +496,20 @@ fn benchmark_add_remove_1_comp_batch_1_000_000(
     bencher.iter[bench_fn]()
 
 
+fn prevent_inlining_add_remove_1_comp_batch_1_000_000() raises:
```
samufi, line 499: Regarding the name (and same below) refer to my earlier comment.

benchmark/world_benchmark.mojo
```diff
@@ -526,6 +536,24 @@ fn benchmark_add_remove_1_comp_1_000_batch_1_000(
...
 
+fn prevent_inlining_add_remove_1_comp_1_000_batch_1_000() raises:
+    pos = Position(1.0, 2.0)
+    comp1 = FlexibleComponent[1](1.0, 42.0)
+    world = SmallWorld()
+
+    # create 1_000 entities that initially do not have FlexibleComponent[1]
+    _ = world.add_entities(pos, count=1_000)
+    # then 1_000 x add component and remove it afterwards
+    for _ in range(1000):
```
samufi, line 547: no for loop required to prevent inlining

benchmark/world_benchmark.mojo
```diff
@@ -500,6 +496,20 @@ fn benchmark_add_remove_1_comp_batch_1_000_000(
     bencher.iter[bench_fn]()
 
 
+fn prevent_inlining_add_remove_1_comp_batch_1_000_000() raises:
+    pos = Position(1.0, 2.0)
+    comp = FlexibleComponent[1](1, 42.0)
+    world = SmallWorld()
+
+    # create 1_000_000 entities that initially do not have FlexibleComponent[1]
+    _ = world.add_entities(pos, count=1_000_000)
```
samufi, line 505: count can be any value, so use something that is fast (1)

benchmark/world_benchmark.mojo
```diff
@@ -606,6 +654,50 @@ fn benchmark_add_remove_5_comp_batch_1_000_000(
...
+    pos = Position(1.0, 2.0)
+    comp1 = FlexibleComponent[1](1.0, 42.0)
+    comp2 = FlexibleComponent[2](2.0, 42.0)
+    comp3 = FlexibleComponent[3](3.0, 42.0)
+    comp4 = FlexibleComponent[4](4.0, 42.0)
+    comp5 = FlexibleComponent[5](5.0, 42.0)
+    world = SmallWorld()
+
+    # create 1_000_000 entities that initially do not have FlexibleComponent[1]
+    _ = world.add_entities(pos, count=1_000_000)
```
samufi, line 667: For inlining prevention, the dynamic values don't matter. Use count=1.

benchmark/world_benchmark.mojo
```diff
@@ -659,14 +751,49 @@ fn benchmark_add_remove_5_comp_1_000_batch_1_000(
...
     world = SmallWorld()
-    entity = world.add_entity(pos)
-    _ = world.replace[Position]().by(entity, vel)
-    query = world.query[Position]().without[Velocity]()
-    _ = world.replace[Position]().by(query, vel)
+
+    # create 1_000 entities that initially do not have FlexibleComponent[1...5]
+    _ = world.add_entities(pos, count=1_000)
+    # then 1_000 x add components and remove them afterwards
+    for _ in range(1000):
```
samufi, line 766: No for loop please

benchmark/world_benchmark.mojo
```diff
@@ -708,6 +835,16 @@ fn benchmark_replace_1_comp_batch_1_000_000(
     bencher.iter[bench_fn]()
 
 
+fn prevent_inlining_batch_replace() raises:
+    world = FullWorld()
+    _ = world.add_entities(FlexibleComponent[0](1.0, 2.0), count=1_000_000)
```
samufi, line 840: See above

benchmark/world_benchmark.mojo
```diff
@@ -782,6 +934,41 @@ fn benchmark_replace_5_comp_1_000_000(
     bencher.iter[bench_fn]()
 
 
+fn prevent_inlining_5_replace() raises:
+    for _ in range(50):
+        world = FullWorld()
+        entities = List[Entity]()
+        for _ in range(1000):
```
samufi, line 941: use minimal entities

benchmark/world_benchmark.mojo
```diff
@@ -822,6 +1009,39 @@ fn benchmark_replace_5_comp_batch_1_000_000(
...
 
+fn prevent_inlining_5_batch_replace() raises:
+    world = FullWorld()
+    _ = world.add_entities(
+        FlexibleComponent[0](1.0, 2.0),
+        FlexibleComponent[1](3.0, 4.0),
+        FlexibleComponent[2](5.0, 6.0),
+        FlexibleComponent[3](7.0, 8.0),
+        FlexibleComponent[4](9.0, 10.0),
+        count=1_000_000,
```
samufi, line 1020: as above.

samufi (changes requested):

samufi (commented):
Overall very nice work. The prevent_inline in the benchmarks should be simplified a bit, and I think many of the filter methods should not return copies or be named differently. As I think you implement stuff that is on the roadmap, the corresponding document could be changed as well.

PR #122, opened 2025-10-24
Test-wise benchmark of position velocity example

samufi:


moseschmiedel:
@samufi Can we close this?

samufi:
I think we can. I guess the initial idea was to come up with benchmarks for comparison with other ECS, but I think this might not be too useful, as the example may be too simple to be representative of real-world scenarios.

moseschmiedel:
Maybe we can look at other ECS implementations and pick something they already implemented as example to compare Larecs to them.

PR #126, opened 2026-04-16
Upgrade to Mojo 1.0.0b2-nightly

### LINKED ISSUE #125: Upgrade to Mojo 0.26.3-nightly
Upgrade to current nightly version of Mojo and eliminate all errors and warnings.

## Subtasks
- [ ] Update the CI testing pipeline.
- [ ] New `modo` version.
- [ ] Updated `StaticOptional`
- [ ] Cleanup
  - [ ] Variadics
    - [ ] Remove explicit variadics overloads
    - [ ] Use `SomeTypeList` where applicable
  - [ ] Unsafe storage of components
  - [ ] Indexing → use `Int`
  - [ ] Bounds checking via assert tool in stdlib.
  - [ ] Constraints via comptime assert or where clauses
  - [ ] Use `InlineArray` instead of `SIMD` for data storage unless SIMD operations are really necessary.
  - [ ] Use conditional conformance where applicable (copyable, writable)
  - [ ] Use default trait implementations and remove unneeded defaults (especially for copy and move, but also in the systems)
  - [ ] Use generic raises on closures in apply (possible via `raises Never`)

moseschmiedel:
Closes #125


src/larecs/unsafe_box.mojo
```diff
@@ -126,41 +134,44 @@ struct UnsafeBox(Copyable, Movable):
...
+            ptr = Optional[UnsafePointer[T, MutExternalOrigin]]()
         else:
-            ptr = UnsafePointer[T].alloc(1)
-            ptr.init_pointee_move(data^)
+            ptr = alloc[T](1)
+            ptr.unsafe_value().init_pointee_move(data^)
 
-        self._data = ptr.bitcast[Byte]()
+        self._data = ptr.unsafe_value().bitcast[Byte]()
         self._destructor = _destructor[T]
```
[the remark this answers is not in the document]
moseschmiedel: `ptr` is of type `Optional[UnsafePointer[T, MutExternalOrigin]]`. Because `Optional` has an implicit `__init__` for constructing it from a value of the wrapped type `alloc[T](1)` constructs an `Optional` implicitly which is then later accessed through `ptr.unsafe_value()`.

src/larecs/pool.mojo
```diff
@@ -58,102 +59,138 @@ struct EntityPool(Copyable, Movable, Sized):
...
         self._available -= 1
-        return self._bits[index(curr)]
 
-    fn _get_new(mut self) raises -> UInt8:
-        """Allocates and returns a new bit. For internal use.
+    def _get_new(mut self, out bit_idx: Int) raises:
+        """Allocates and returns a new bit index.
+
+        This internal helper only allocates fresh indices; callers should use
+        [.BitPool.get] so recycled indices are preferred.
```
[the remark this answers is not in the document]
moseschmiedel: `def foo(out <name>: <type>)` is just another syntax for writing `def foo() -> <type>`.

src/larecs/unsafe_box.mojo
```diff
@@ -126,41 +134,44 @@ struct UnsafeBox(Copyable, Movable):
...
         Destructor for the UnsafeBox.
 
         It destroys the data stored in the box and frees the memory
         associated with the box.
         """
+        debug_assert(
+            self._data is not None,
+            "Attempting to destroy an empty UnsafeBox.",
+        )
         self._destructor(self._data)
```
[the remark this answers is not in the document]
moseschmiedel: We require `UnsafeBox` to always point to non-zero-sized T value.

.github/workflows/main.yml
```diff
@@ -1,48 +1,47 @@
...
+        run: |
+          curl -fsSL https://pixi.sh/install.sh | sh
+
+      - name: Install mojo and dependencies
+        run: |
+          pixi install --locked
+
+      - name: Run tests
+        run: |
+          scripts/run_tests.sh test
```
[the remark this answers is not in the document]
moseschmiedel: The executable bit is set in git!

benchmark/plots/pixi.toml
```diff
@@ -1,15 +1,15 @@
...
 name = "plots"
-platforms = ["linux-64"]
+platforms = ["linux-64", "osx-arm64"]
 version = "0.4.0"
 
 [tasks]
 
 [dependencies]
-mojo = "==0.25.6"
+mojo = ">=0.26.2, <1.0.0"
```
samufi: The PR says you are upgrading to 1.0.0b2. Wouldn't it make sense to fix the version here (and switch to the stable branch)?

benchmark/custom_benchmark.mojo
```diff
@@ -1,27 +1,29 @@
...
 
-fn DefaultBench() raises -> Bench_:
+def DefaultBench() raises -> Bench_:
     """Returns the default benchmarking struct."""
     return Bench_(DefaultConfig())
 
 
 @fieldwise_init
-struct Bencher:
+struct Bencher(ImplicitlyCopyable):
```
samufi: We might currently only use DefaultBench. Maybe we should remove the other parts of this file? Or do you intend to use them later?

scripts/run_tests.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
samufi: We may take inspiration from the model-croe test script that also allows asan and debug info. This can come in handy if nomething goes wrong.

AGENTS.md
```diff
@@ -1,32 +1,48 @@
...
+- **Systems**: Logic that operates on entities with specific components
+- **World**: Central container managing entities, components, and systems
+- **Archetypes**: Efficient storage for entities with the same component composition
+- **Queries**: Fast iteration over entities matching specific criteria
+
 ## Build/Test Commands
 
-- Run all tests: `pixi run mojo test -I src/ test/`
-- Run single test: `pixi run mojo test -I src/ test/<filename>.mojo`
+- Run all tests: `pixi run mojo run -D ASSERT=all -I src/ test/_all.mojo`
```
samufi: This requires an explicit opt-in. Wouldn't it be better to use the testing script? One less thing that can break. I sugget removing `test/_all.mojo`.

src/larecs/_utils.mojo
```diff
@@ -1,103 +1,74 @@
...
+    b: InlineArray[ElementType, b_size],
+    out result: InlineArray[ElementType, a_size + b_size],
+):
+    result = {uninitialized = True}
 
-    Returns:
-        The next power of two greater than or equal to the given value.
-    """
-    return UInt(next_pow2[DType.uint](value))
+    memcpy(dest=result.unsafe_ptr(), src=a.unsafe_ptr(), count=a_size)
```
samufi: Would it make sense to use https://mojolang.org/docs/std/memory/memory/uninit_move_n/ and friends here? The current implementation seems unsafe to me if the move constructor is not trivial.

src/larecs/_utils.mojo
```diff
@@ -1,103 +1,74 @@
...
-    - **Fragmentation Reduction**: Reduces memory fragmentation in long-running systems
-
-    **Mathematical Properties:**
-    - `next_pow2(0)` returns 1 (special case for zero input)
-    - `next_pow2(n)` returns `n` if `n` is already a power of 2
-    - `next_pow2(n)` returns the smallest `2^k` where `2^k >= n`
 
-    Parameters:
-        dtype: The scalar data type. Must be an integral type for bit operations.
+def _assert_index_in_bounds(index: Int, size: Int):
```
samufi: We have https://mojolang.org/docs/std/collections/check_bounds/check_bounds/ now. Can we use that?

src/larecs/_utils.mojo
```diff
@@ -1,103 +1,74 @@
...
 
-    Returns:
-        The next power of two greater than or equal to the given value.
 
-    **Reference:**
-    Based on the bit-smearing algorithm from Stanford Graphics Lab's bit manipulation
-    reference: https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2.
-    """
-    constrained[dtype.is_integral(), "expected integral dtype"]()
+def _assert_range_in_bounds(start_index: Int, count: Int, size: Int):
```
samufi, line 54: If we use the builtin check_bounds, align the implementation.

src/larecs/archetype.mojo
```diff
@@ -121,203 +149,578 @@ struct EntityAccessor[
...
+
+
+struct _ComponentStorage[*ComponentTypes: ComponentType](
+    Copyable, ImplicitlyDestructible, Movable, Sized
+):
     """
-    Archetype represents an ECS archetype.
+    Internal struct to store component data for an archetype.
+
+    UnsafePointers to the component buffers are stored in a sparse tuple indexed by component ID (position in the ComponentTypes TypeList).
```
samufi, line 183: Are the unused pointers really dangling? (I mean we use Optional - so maybe it is better to say they are "None" or "inactive")

src/larecs/archetype.mojo
```diff
@@ -121,203 +149,578 @@ struct EntityAccessor[
...
-        component_manager: The component manager.
+        ComponentTypes: The component types of the world.
     """
 
-    alias dType = BitMask.IndexDType
-    """The DType of the component ids."""
-
-    alias Id = SIMD[Self.dType, 1]
-    """The type of the component ids."""
+    comptime component_manager = ComponentManager[*Self.ComponentTypes]
```
samufi, line 199: Please use docstrings for all members.

src/larecs/archetype.mojo
```diff
@@ -121,203 +149,578 @@ struct EntityAccessor[
...
+        self.capacity = copy.capacity
+        self.size = copy.size
+        self.active_component_mask = copy.active_component_mask
+        self.data = Self._PointerTuple()
 
-    # Index of the the archetype's node in the archetype graph
-    var _node_index: UInt
+        def copy_component_ptr[
+            T: ComponentType, id: ComponentId
+        ](ptr: Self.ComponentPointer[T]) capturing -> Self.ComponentPointer[T]:
```
samufi: Use unified closure here if possible.

src/larecs/archetype.mojo
```diff
@@ -121,203 +149,578 @@ struct EntityAccessor[
...
+    def __len__(self) -> Int:
+        """
+        Gets the current size (number of stored entities) of the component storage.
 
         Returns:
-            The zero archetype.
+            The current size of the component storage.
         """
-        self = Self.__init__[used_internally=True](0, BitMask(), 0)
+        result = self.size
```
samufi: return the size directly; make this `@always_inline`

src/larecs/archetype.mojo
```diff
@@ -121,203 +149,578 @@ struct EntityAccessor[
...
-        *, used_internally: Bool
-    ](out self, node_index: UInt, mask: BitMask, capacity: UInt):
-        """Initializes the archetype without allocating memory for components.
+        self._for_active_components[
+            is_mutating=False, func=free_component_storage
+        ]()
 
-        Note:
-            Do not use this constructor directly!
+    def copy_deep(self, out new_storage: Self):
```
samufi: I think `deep_copy()` would be a more common name for this. (Otherwise `copy_deeply`). I would typically expect `copy` to always do a deep copy and make the shallow copy the special case.

src/larecs/archetype.mojo
```diff
@@ -121,203 +149,578 @@ struct EntityAccessor[
...
+
+        new_storage._for_active_components[
+            is_mutating=True, func=memcpy_component
+        ]()
+
+    def unsafe_init_components(mut self, read init_component_mask: BitMask):
+        """(Re)Initializes owned component storage while keeping the component layout intact.
+
+        Important:
+        This is intended for internal ownership-transfer flows where another archetype has
```
samufi, line 328: If this is intended for internal use, make it a `_under` method.

src/larecs/archetype.mojo
```diff
@@ -121,203 +149,578 @@ struct EntityAccessor[
...
+            if init_component_mask.get(id):
+                return rebind[Self.ComponentPointer[T]](
+                    Optional(alloc[T](self.capacity))
+                )
+            else:
+                return ptr
+
+        self._for_active_components[is_mutating=True, func=init_component_ptr]()
+
+    def get_component_count(self) -> Int:
```
samufi, line 358: Use a direct return and make it `@always_inline`

src/larecs/archetype.mojo
```diff
@@ -121,203 +149,578 @@ struct EntityAccessor[
...
+        Returns:
+            The index of the newly added entity.
+        """
+        if self.size == self.capacity:
+            self.reserve(max(self.capacity * 2, 8))
+        var idx = self.size
+        self.size += 1
+        return idx
+
+    def clear(mut self):
```
samufi, line 380: Make all trivial functions and all that are almost trivial but potentially part of a fast path `@always_inline`

src/larecs/archetype.mojo
```diff
@@ -121,203 +149,578 @@ struct EntityAccessor[
...
+
+        self.size -= 1
+
+        if remove_idx == self.size:
+            return False
+
+        def swap_component_data[
+            T: ComponentType, id: ComponentId
+        ](ptr: UnsafePointer[T, MutExternalOrigin]) capturing:
+            memcpy(
```
samufi: Use https://mojolang.org/docs/std/memory/memory/uninit_move_n/

src/larecs/archetype.mojo
```diff
@@ -121,203 +149,578 @@ struct EntityAccessor[
...
-        self = Self(node_index, mask_, component_ids, capacity)
+        self.assert_has_component[T]()
 
-    fn __init__[
-        component_count: Int
+        optional_ptr = rebind[Self.ComponentPointer[T]](self.data[id])
+        assert (
+            optional_ptr is not None
+        ), "Active component storage must be initialized"
+        result = optional_ptr.value()
```
samufi: `.value()` already checks if the value exists. The assert above should be unneeded.

samufi: And isn't `assert_has_component` already doing the check?

src/larecs/archetype.mojo
```diff
@@ -121,203 +149,578 @@ struct EntityAccessor[
...
+            count: The number of elements to set.
+            from_idx: The index of the first entity in the storage to copy from.
+        """
+        _assert_range_in_bounds(to_idx, count, self.size)
+        _assert_range_in_bounds(from_idx, count, storage.size)
+
+        if count == 0:
+            return
+
+        memcpy(
```
samufi: See comments for the other memcps. Maybe we should do a search replace.

src/larecs/archetype.mojo
```diff
@@ -121,203 +149,578 @@ struct EntityAccessor[
...
+                    dest=self.get_component_ptr[T]() + to_idx,
+                    src=source[].get_component_ptr[T]() + from_idx,
+                    count=count,
+                )
+
+    def _for_active_components[
+        is_mutating: Bool,
+        func: def[T: ComponentType, id: ComponentId](
+            ptr: Self.ComponentPointer[T]
+        ) capturing -> Self.ComponentPointer[T],
```
samufi: As commented above, it may be good to use the new closure type here.

src/larecs/archetype.mojo
```diff
@@ -472,303 +892,301 @@ struct Archetype[
...
         )
 
-    @always_inline
-    fn unsafe_set[
-        T: Indexer
-    ](mut self, idx: T, id: Self.Id, value: UnsafePointer[UInt8]):
-        """Sets the component with the given id at the given index.
+        _trace_function["OUT"]("Archetype.get_entity_accessor")
+
+    #
```
samufi: Will this be used in future?
If so, add a comment. If not, delete.

src/larecs/component.mojo
```diff
@@ -1,161 +1,117 @@
...
 
     @staticmethod
     @always_inline
-    fn get_id[T: ComponentType]() -> Self.Id:
+    def get_id[T: ComponentType]() -> Self.Id:
         """Get the ID of a component type.
 
         Parameters:
-            T: The component type.
+            T: The component type. Constraints: Must be in the list of component types.
```
samufi, line 93: Add a section for constraints

src/larecs/constants.mojo
```diff
@@ -1,10 +1,10 @@
 # Maximal UInt16 number
-alias MAX_UINT16 = 2**16 - 1
+comptime MAX_UINT16 = 2**16 - 1
```
samufi: This is the only constant here that is actually used. This can also be obtained from the stdlib. I suggest removing this module entirely.

src/larecs/query.mojo
```diff
@@ -67,57 +130,104 @@ struct Query[
...
+        it = ArchetypeByMaskIterator(
+            mask_iterator=ArchetypeByMaskIterator.mask_iterator(
+                Pointer(to=self._world[]._archetypes),
+                self._mask,
+                self._without_mask.copy(),
+            )
         )
 
+        try:
+            iterator = Self.World.Iterator[
```
samufi: Use an initializer list here to reduce verbosity

src/larecs/world.mojo
```diff
@@ -15,52 +18,122 @@ from .component import (
...
+    comptime missing_components_for_removal_query = WorldError(_variant=4)
+    comptime duplicate_components_for_addition_entity = WorldError(_variant=5)
+    comptime duplicate_components_for_addition_query = WorldError(
+        _variant=6,
     )
+    comptime out_of_locks = WorldError(_variant=7)
+    comptime unbalanced_unlock = WorldError(_variant=7)
+
+    def variant_name(self) -> String:
+        if self._variant == Self.non_existent_entity._variant:
```
samufi: Would it be smarter / more readable to use a dict here?

test/_all.mojo
```diff
@@ -0,0 +1,38 @@
+from std.testing import TestSuite
```
samufi: As noted already elsewhere, an automated testing script may be safer than an opt-in implementation.

test/scheduler_test.mojo
```diff
@@ -44,7 +74,8 @@ struct MeanState(ResourceType):
 #         6,
 #     )
```
samufi: I suggest uncommenting the test below. I think it was disabled before due to a Mojo bug and should work now. You may use the updated code below
```suggestion
@fieldwise_init
struct TestSystem[copies: Int, count: Int = 10](System):
    var a: Int

    def __init__(out self):
        self.a = 0

    def initialize(mut self, mut world: World) raises:
        assert_equal(self.a, 0)
        _ = world.add_entities(self.a, count=self.count)
        self.a = 1

    def update(mut self, mut world: World) raises:
        assert_equal(self.a, 1)
        assert_equal(len(world), self.count * self.copies)
        for entity in world.query[Int]():
            entity.get[Int]() += 1

    def finalize(mut self, mut world: World) raises:
        sum = 0
        var counter = 0
        for entity in world.query[Int]():
            sum += entity.get[Int]()
            counter += 1
        world.resources.set[add_if_not_found=True](MeanState(Float64(sum) / Float64(counter)))


def test_scheduler() raises:
    scheduler = Scheduler[Int, Float64]()
    scheduler.add_system(TestSystem[2]())
    scheduler.add_system(TestSystem[2]())
    scheduler.run(3)
    assert_equal(
        scheduler.world.resources.get[MeanState]().value,
        6,
    )

comptime functions = __functions_in_module()
```

samufi: If you think the test is unnecessary, delete the commented code.

samufi (commented):
Thanks, looks good! 

Besides some minor comments, one question: do we want to make larecs work on non-trivial structs here already? In any event, I think it would be good to replace the memcps etc. with the more general versions from the stdlib (which call the move constructor if necessary). Howeer, we can also do this in a later PR.

src/larecs/_utils.mojo
```diff
@@ -1,103 +1,74 @@
...
 @always_inline
-fn concatenate_inline_arrays[
-    ElementType: Copyable & Movable, a_size: Int, b_size: Int
-](
-    a: InlineArray[ElementType, a_size],
-    b: InlineArray[ElementType, b_size],
-    out result: InlineArray[ElementType, a_size + b_size],
-):
-    result = {uninitialized = True}
+def _trace_function[inout: StaticString](name: StaticString):
```
samufi: I think we should use a context manager for this. This is much cleaner and safer. And maybe let it take a function as parameter and use reflection to extract the name. This gives us additional security that the function exists.

samufi: Maybe it makes sense to even put this into a dedicated module for performance measuring / tracing.

src/larecs/_utils.mojo
moseschmiedel: No, currently there is a [bug during debug builds](https://github.com/modular/modular/issues/6636). We have to wait for the fix to remove this.

src/larecs/component.mojo
moseschmiedel: According to the [Mojo docstring style guide](https://github.com/modular/modular/blob/main/mojo/stdlib/docs/docstring-style-guide.md#constraints) this should be the desirable format.

src/larecs/world.mojo
moseschmiedel: I used an `InlineArray` with `global_constant` for this.

src/larecs/_utils.mojo
moseschmiedel: > And maybe let it take a function as parameter and use reflection to extract the name. This gives us additional security that the function exists.

This does not work on generic functions at the moment because there is no ergonomic way to distinguish between variants of overloaded functions 🙁

src/larecs/archetype.mojo
samufi: And for all aliases. If you do not agree, let us talk, but I think this is a good general standard to have.

src/larecs/archetype.mojo
```diff
@@ -121,203 +142,617 @@ struct EntityAccessor[
...
+            storage_size: Int,
+            storage_capacity: Int,
+            comp_ptr: UnsafePointer[T, MutExternalOrigin],
+        ):
+            destroy_n(comp_ptr, count=storage_size)
+            comp_ptr.free()
+
+        self._for_active_components[is_mutating=False,](free_component_storage)
+
+    def shallow_copy(self, out new_storage: Self):
```
samufi, line 306: In what sense is this different from a move? The underlying data are remaining untouched, right?

src/larecs/archetype.mojo
```diff
@@ -121,203 +142,617 @@ struct EntityAccessor[
...
+            storage_size: Int,
+            storage_capacity: Int,
+            comp_ptr: UnsafePointer[T, MutExternalOrigin],
+        ),
+    ](self, func: FuncType) where not is_mutating:
+        """Helper to iterate over active components and their pointers.
+
+        Parameters:
+            is_mutating: Whether the function mutates the component pointers. This determines whether the pointers passed to the function are mutable or not.
+            func: A function that takes a component ID and the corresponding typed pointer, and performs some operation.
```
samufi, line 712: Unknown Parameter func.
Document both the parameters and the arguments

src/larecs/_tracing.mojo
```diff
@@ -0,0 +1,43 @@
...
+@fieldwise_init
+struct TraceGuard(ImplicitlyCopyable):
+    """A guard object for function tracing.
+
+    When created, it emits an "IN" trace, and when dropped, it emits an "OUT" trace.
+    """
+
+    var name: StaticString
+
+    def __enter__(mut self):
```
samufi, line 39: Make the methods @always_inline and add docstrings for the members.

src/larecs/component.mojo
samufi: Thanks. I was not aware and will need to revisit some of the stuff I did myself.

samufi:
Thanks, some additional comments. I have now looked at your changes in response to my comments. When you have addressed the comments, I will do another quick check of the entire PR.

samufi (changes requested):
See my previous comment.

src/larecs/archetype.mojo
moseschmiedel: In a move the metadata would not be copied but moved elsewhere. This is therefore not a move because after this we have to `_ComponentStorage` instances using the same data (think of it as a pointer).

src/larecs/archetype.mojo
```diff
@@ -121,203 +145,645 @@ struct EntityAccessor[
...
+        Args:
+            func: A function that takes a component ID and the corresponding typed pointer, and performs some operation.
+                If `is_mutating` is True, the function can return a new pointer to replace the existing one in the storage (e.g. for reallocations), which will be updated accordingly.
+                If `is_mutating` is False, the function must not return anything and the pointers passed to it must not be modified.
+        """
+        comptime for id in range(len(Self.ComponentTypes)):
+            comptime T = Self.ComponentTypes[id]
+            if self.has_component[T]():
+                comp_ptr = rebind[Self.ComponentPointer[T]](self.data[id])
+                assert (
```
samufi: Better debug_assert? Or is this never in a hot loop? (I'd be also okay with leaving this as is, just want it to be a concious decision)

src/larecs/archetype.mojo
```diff
@@ -121,203 +145,645 @@ struct EntityAccessor[
...
+
+    def _for_active_components[
+        is_mutating: Bool,
+        FuncType: def[T: ComponentType, id: ComponentId](
+            storage_size: Int,
+            storage_capacity: Int,
+            comp_ptr: UnsafePointer[T, MutExternalOrigin],
+        ),
+    ](self, func: FuncType) where not is_mutating:
+        """Helper to iterate over active components and their pointers.
```
samufi: The two overloads of `_for_active_components` are doing distinct things. I would therefore also use distinct documentation. In this version, for example, `is_mutating` is always `False`, so I think we do not need to document `func` for the reverse case. Instead, do this in the method above. I think it is confusion to see two methods with the same docs that do different things.

src/larecs/archetype.mojo
```diff
@@ -121,203 +145,645 @@ struct EntityAccessor[
...
+        """
+        comptime for id in range(len(Self.ComponentTypes)):
+            comptime T = Self.ComponentTypes[id]
+            if self.has_component[T]():
+                comp_ptr = rebind[Self.ComponentPointer[T]](self.data[id])
+                self.data[id] = rebind[Self._PointerTuple.element_types[id]](
+                    func[T, id](self.size, self.capacity, comp_ptr)
+                )
+
+    def _for_active_components[
```
samufi: Would it be better to call this `_apply_to_active_components` in line with other `apply` API? Also, be more elaborate in the doc string. This is not only a helper for iterating over the components but also to apply some function to them.

src/larecs/archetype.mojo
```diff
@@ -121,203 +145,645 @@ struct EntityAccessor[
...
+            T: ComponentType, id: ComponentId
+        ](
+            storage_size: Int,
+            storage_capacity: Int,
+            comp_ptr: UnsafePointer[T, MutExternalOrigin],
+        ):
+            destroy_n(comp_ptr, count=storage_size)
+            comp_ptr.free()
+
+        self._for_active_components[is_mutating=False,](free_component_storage)
```
samufi:
```suggestion
        self._for_active_components[is_mutating=False](free_component_storage)
```

src/larecs/archetype.mojo
```diff
@@ -121,203 +145,645 @@ struct EntityAccessor[
...
+                If `is_mutating` is False, the function must not return anything and the pointers passed to it must not be modified.
+        """
+        comptime for id in range(len(Self.ComponentTypes)):
+            comptime T = Self.ComponentTypes[id]
+            if self.has_component[T]():
+                comp_ptr = rebind[Self.ComponentPointer[T]](self.data[id])
+                assert (
+                    comp_ptr is not None
+                ), "Active component storage must be initialized"
+                func[T, id](self.size, self.capacity, comp_ptr.value())
```
samufi: `Optional.value()` aborts on empty optionals. If this is acceptable here, we can get rid of the assert above.

src/larecs/archetype.mojo
```diff
@@ -121,203 +145,645 @@ struct EntityAccessor[
...
-    var _entities: List[Entity]
+    comptime _PointerTuple = Tuple[
+        *Self.ComponentTypes.map[Self._PointerMapper]()
+    ]
+    """The type of the tuple storing component pointers for all component types. See the description of [._ComponentStorage] for the layout and semantics of this tuple.
+    """
 
-    # Index of the the archetype's node in the archetype graph
-    var _node_index: UInt
+    var capacity: Int
```
samufi: Shouldn't these attributes be private? When they are changed from the outside, the instance may get inconsistent.

samufi (changes requested):
Hi, I now went another time through the relevant parts. I have some last comments. When they are fixed, we are ready to merge. Thanks for your awesome work!

src/larecs/archetype.mojo
```diff
@@ -121,203 +145,634 @@ struct EntityAccessor[
...
+    var _capacity: Int
+    """The capacity of the component storage, i.e. how many entities can be stored without reallocating."""
+    var _size: Int
+    """The current size of the component storage, i.e. how many entities are currently stored."""
+    var data: Self._PointerTuple
+    """The component data, stored as typed pointers to component buffers."""
 
-    # Mask of the the archetype's node in the archetype graph
-    var _mask: BitMask
+    var active_component_mask: BitMask
```
samufi: Make this private

src/larecs/archetype.mojo
```diff
@@ -121,203 +145,634 @@ struct EntityAccessor[
...
+    """The type of the tuple storing component pointers for all component types. See the description of [._ComponentStorage] for the layout and semantics of this tuple.
+    """
 
-    # Index of the the archetype's node in the archetype graph
-    var _node_index: UInt
+    var _capacity: Int
+    """The capacity of the component storage, i.e. how many entities can be stored without reallocating."""
+    var _size: Int
+    """The current size of the component storage, i.e. how many entities are currently stored."""
+    var data: Self._PointerTuple
```
samufi: Make this private

samufi (changes requested):

src/larecs/archetype.mojo
```diff
@@ -29,89 +41,101 @@ struct EntityAccessor[
...
-        @parameter
-        for i in range(components.__len__()):
-            self._archetype[].get_component[T = Ts[i]](
-                self._index_in_archetype
-            ) = components[i].copy()
+        with TraceGuard(name="EntityAccessor.set"):
+            comptime assert constrain_components_unique[
+                *Ts
+            ](), "Component types must be unique."
+            # comptime assert Self.component_manager._ContainsComponents[
```
samufi: Can this be removed?

samufi (approved):

Issue #128, opened 2026-04-24
Rework ArchetypeIterators once trait parameters are there


moseschmiedel:
Currently `archetype.mojo` has some complicated structs, so that `_EntityIterator` can accept both an `ArchetypeByMaskIterator` and `ArchetypeByListIterator`. Once trait parameters are available we should be able to delete lots of code here and still achieve Polymorphy in `_EntityIterator`.

moseschmiedel:
There is now only one `ArchetypeIterator`. So this is not needed anymore.

PR #131, opened 2026-05-13
Unify iterators for `Archetypes` and `Entities` of a `World` 

### LINKED ISSUE #26: Segfault when world is in "distant" context
The following code leads to a seg fault:

```python
fn benchmark_query_1_comp_1_000_000(
    inout bencher: Bencher,
) raises capturing:
    pos = Position(1.0, 2.0)
    world = World[Position, Velocity]()

    @always_inline
    @parameter
    fn bench_fn() capturing raises:
        for _ in range(1000):
            _ = world.new_entity(pos)

    bencher.iter[bench_fn]()
```

However, the following works:
```python
fn benchmark_query_1_comp_1_000_000(
    inout bencher: Bencher,
) raises capturing:
    pos = Position(1.0, 2.0)

    @always_inline
    @parameter
    fn bench_fn() capturing raises:
        world = World[Position, Velocity]()
        for _ in range(1000):
            _ = world.new_entity(pos)

    bencher.iter[bench_fn]()
```

It is unclear to me what causes the issue and whether this is a problem on the mojo or the larecs side.

### LINKED ISSUE #127: Use typed errors
See `world.mojo` for reference.

### LINKED ISSUE #136: Reenable move ctor tests when stdlib is fixed
Currently four move ctor tests are disabled due to a [bug in the Mojo stdlib](https://github.com/modular/modular/issues/6682).

List of disabled tests:
- `test/unsafe_box_test.mojo`:
  - `test_unsafe_box_copy_move_del()`
- `test/static_optional_test.mojo`:
  - `test_comptime_optional_move_del()`
- `test/archetype_test.mojo`:
  - `test_archetype_reserve_non_trivial_component()`
  - `test_archetype_remove_non_trivial_component()`

Reenable them once the upstream issue is fixed!

### LINKED ISSUE #139: Change `Query.{without,exclusive}()` to consume self
Optimize `Query.{without,exclusive}()` methods so that they perform an in place operation. For this `QueryInfo` should be extended with both an `without` and `exclusive` method. `Query` can then use these to perform the in place operation and just change the stored `_info` field.

### Example usage
```mojo
query1 = world.query[Position]().without[Velocity]() # same as before

query2 = world.query[Position]()
query3 = query2^.without[Velocity]() # query2 needs to be moved when using without and is not usable anymore

# following line should fail at comptime
for entity in query2:
    ...

query4 = query3.copy().without[Velocity]() # query3 is explicitly copied so that it still can be used later on
query5 = query4.without[Velocity]() # same as query4 with implicit copying (I think...)

# still usable
for entity in query3:
    ...
for entity in query4:
    ...
```

moseschmiedel:


src/larecs/query.mojo
```diff
@@ -1147,19 +852,30 @@ struct _EntityIterator[
             Whether there are more elements to iterate.
         """
         _trace_function["IN"]("_EntityIterator.__has_next__")
-        result = self._entity_index < self._last_entity_index
+        if (
```
samufi: Do we still need `__has_next__` if we use the raising mechanism? I believe hasnext will be deprecated soon, so if we do not need this for logical reasons, we should remove it. Same applies to the other iterators.

samufi: I now see that we may use it within `__bool__`. Nonetheless, I think if we do not use the dunder approach, we should turn it into a "usual" private function.

src/larecs/world.mojo
```diff
@@ -1960,43 +1872,37 @@ struct World[*component_types: ComponentType](Copyable, Movable, Sized):
...
-        _trace_function["OUT"]("World._get_entity_iterator")
+        finally:
+            _trace_function["OUT"]("World._get_entity_iterator")
 
     @always_inline
     def _get_archetype_iterator[
         has_without_mask: Bool = False
     ](
-        ref self,
+        mut self,
```
samufi: Why does this need to be `mut`?

samufi (changes requested):
I think we should unifiy the mechanism for stopping an iteration and always use the raise StopIteration method. The old has_next mechanism should be removed. Otherwise fine!

src/larecs/world.mojo
moseschmiedel: Changed the `_ArchetypeIterator` construction to not need this `mut` here anymore.

moseschmiedel:
Resolves #128 as there is now only one ArchetypeIterator

benchmark/world/component_benchmark.mojo
```diff
@@ -0,0 +1,31 @@
+import world_component_single_benchmark
+import world_component_single_batch_benchmark
+import world_component_multi_benchmark
+import world_component_multi_batch_benchmark
```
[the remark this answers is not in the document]
moseschmiedel: Removed this as it's not needed

benchmark/world/component_benchmark.mojo
```diff
@@ -0,0 +1,31 @@
...
+    )
+    world_component_single_batch_benchmark.run_all_world_component_single_batch_benchmarks(
+        bench
+    )
+    world_component_multi_benchmark.run_all_world_component_multi_benchmarks(
+        bench
+    )
+    world_component_multi_batch_benchmark.run_all_world_component_multi_batch_benchmarks(
+        bench
+    )
```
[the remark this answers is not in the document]
moseschmiedel: I removed the file as we don't need it.

benchmark/world/replace_single_benchmark.mojo
```diff
@@ -0,0 +1,59 @@
...
+
+def run_all_world_replace_single_benchmarks(mut bench: Bench) raises:
+    bench.bench_function(
+        benchmark_replace_1_comp_1_000_000,
+        BenchId("10^6 * replace 1 component"),
+    )
+    prevent_inlining_replace()
+
+
+def main() raises:
```
samufi: main is not supported in a package. It also is unneeded. Remove here and elsewhere.

src/larecs/query.mojo
```diff
@@ -774,45 +619,52 @@ struct ArchetypeIterator[
...
+            if not self._has_next():
+                raise StopIteration()
+            accessor = self.archetype[].get_entity_accessor(
+                self._index,
+            )
+            self._index += 1
 
 
-struct _EntityIterator[
+struct _WorldIterator[
```
samufi: If an ArchetypeIterator iterates over archetypes and an EntityIterator over entities, WorldIterator may be a confusing name. Maybe `WorldEntityIterator` because it iterates over the entities in a world? (And then `ArchetypeEntityIterator` instead of the EntityIterator?)

src/larecs/query.mojo
```diff
@@ -933,100 +793,95 @@ struct _EntityIterator[
...
-            debug_assert(
-                Bool(self._current_archetype), "No more archetypes to iterate."
-            )
-            accessor = (
-                self._current_archetype.unsafe_value()[].get_entity_accessor(
-                    self._entity_index,
+        with TraceGuard(name="_WorldIterator.__next__"):
+            if (
+                not self._entity_iterator
+                or not self._entity_iterator.unsafe_value()._has_next()
```
samufi: If we use the private function _has_next here (outside of the struct), should we maybe make it part of the public API? Is this not already done via the bool conversion? Then I think we should use this here (and elsewhere outside of the original struct)

samufi (changes requested):
Looks very good. Very minor comments.

benchmark/world/access_benchmark.mojo
```diff
@@ -0,0 +1,215 @@
...
+    def bench_fn() {read}:
+        try:
+            world = SmallWorld()
+            entity = world.add_entity(c1, c2, c3, c4, c5)
+            for _ in range(500_000):
+                world.set(entity, c1_2, c2_2, c3_2, c4_2, c5_2)
+                world.set(entity, c1, c2, c3, c4, c5)
+
+        except e:
+            print(e)
```
samufi, line 89: The initialization of world can now be moved out of the measured block. I think this PR solves https://github.com/samufi/larecs/issues/26

Suggestion:
```mojo
    try:
        world = SmallWorld()
        entity = world.add_entity(c1, c2, c3, c4, c5)
    except e:
        return
    
    @always_inline
    def bench_fn() {mut world, read}:
        try:
            for _ in range(500_000):
                world.set(entity, c1_2, c2_2, c3_2, c4_2, c5_2)
                world.set(entity, c1, c2, c3, c4, c5)
        except e:
            print(e)

```

samufi: Of course this applies to all benchmarks.

samufi (approved):

samufi (approved):
Lgtm, if you have checked the copilot comments. I am still not completely sure why we need so much biolerplate for this. perf could do profiling without an extra framework. But we can discuss this next coffee break.

PR #146, opened 2026-06-26
Small fixes

samufi:
This PR combines several small fixes that were necessary to use Larecs in a project.

### DIFF
--- a/src/larecs/archetype.mojo
+++ b/src/larecs/archetype.mojo
@@ -1237,8 +1237,7 @@ struct Archetype[
             `count` indices.
         """
         with TraceGuard(name="Archetype.extend"):
-            if count <= 0:
-                return self._storage._size - 1
+            debug_assert(count > 0, "Count must be positive.")
 
             start_index = self._storage._size
 
--- a/src/larecs/component.mojo
+++ b/src/larecs/component.mojo
@@ -42,7 +42,7 @@ def constrain_valid_components[*Ts: ComponentType]() -> Bool:
 
 struct ComponentManager[
     *ComponentTypes: ComponentType,
-](TrivialRegisterPassable):
+](TrivialRegisterPassable, Writable):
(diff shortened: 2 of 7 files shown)


moseschmiedel (approved):
~Looks good.~ Thanks for the fixes.

moseschmiedel:
> Looks good. Thanks for the fixes!

Oops, didn't see the CI failures... Your change in `world.add_entity` leads to a possible raise of `WorldError`. This causes two different `Error` types in a `try / except` block in for example in `query_benchmark.mojo:35` (`Query.__iter__()` raises a standard `Error` and `world.add_entity(pos)` raises `WorldError`). Mojo does not allow this.
Could you either change the relevant benchmarks and have two separate `try / except` blocks or change either one of the Error types on `World.add_entity()` or `Query.__iter__()`

moseschmiedel (changes requested):
Hmm. It's interesting as you didn't touch `World.add_entity()` but only `World.add_entities()`... Maybe for now we make all functions have only a bare `raises` without a type as working with different error types seems difficult atm.

samufi:
> Hmm. It's interesting as you didn't touch `World.add_entity()` but only `World.add_entities()`... Maybe for now we make all functions have only a bare `raises` without a type as working with different error types seems difficult atm.

OK, how much should I change? Only the parts that I have touched? (I was thinking the CI is failing due to the current Mojo version - we could also see if the issue is resolved now / soon)

moseschmiedel:
> OK, how much should I change? Only the parts that I have touched? (I was thinking the CI is failing due to the current Mojo version - we could also see if the issue is resolved now / soon)

I think we should try to keep the CI successful. I'd suggest you change all functions to have a bare `raises`.

moseschmiedel:
There is also an open issue for the typed error handling #127

moseschmiedel (approved):

PR #147, opened 2026-06-29
Tracy

moseschmiedel:
[Tracy](https://wolfpld/tracy) is a cross-platform profiling application which supports various tools to analyze the behavior of Larecs in real-time.
This replaces the old stdout based `_tracing.mojo` approach.


samufi (changes requested):
Ah, one request I do have: could you add a readme entry on how to do profiling? Or at least a link to a plcae where this is explained.

pixi.toml
```diff
@@ -15,22 +15,23 @@ preview = ["pixi-build"]
 tests = { cmd = "test/run_tests.sh", default-environment = "test" }
 benchmarks = { cmd = "mojo run -I src benchmark/run_benchmarks.mojo", default-environment = "default" }
 update-mojo = { cmd = "python scripts/update_mojo.py", default-environment = "default" }
+profiling = { cmd = "mojo run -I src -Xlinker -L\"$CONDA_PREFIX/lib\" -Xlinker -lmojotracy -DTRACY_ENABLE benchmark/run_benchmarks.mojo", default-environment = "default" }
 
 [dependencies]
 mojo = "==1.0.0b3.dev2026062206"
+mojo-tracy = ">=0.2.0"
```
samufi, line 22: could tracy also be a dependency for the profiling job only?

src/larecs/archetype.mojo
```diff
@@ -524,30 +580,37 @@ struct _ComponentStorage[*ComponentTypes: ComponentType](
...
+            comptime assert constrain_components_unique[
+                *Ts
+            ](), "Component types must be unique."
+            _assert_index_in_bounds(entity_idx, self._size)
+
+            Self.component_manager.assert_valid_components[*Ts]()
+            self.assert_has_components[*Ts]()
+
+            @always_inline
+            def set_component[
```
samufi, line 598: I guess the closure is necessary because of the api? I would still use the original set_component function here (inside the closure) to avoid redundancy. It should be inlined anyway.

src/larecs/scheduler.mojo
```diff
@@ -77,8 +88,16 @@ def _initialize_system[
...
-    ref concrete_system = system.unsafe_get[S]()
-    S.initialize[*ComponentTypes](concrete_system, world)
+    with Zone(
+        function_name=(
+            "scheduler._initialize_system[S: System, *ComponentTypes:"
+            " ComponentType](mut system: UnsafeBox, mut world:"
+            " World[*ComponentTypes])"
+        )
+    ):
+        with Zone(function_name=String(t"{reflect[S].name()}.initialize()")):
```
samufi: Why do you nest tracy zones? (Here and below)

samufi (changes requested):
Looks good mostly. Minor requests / questions.

pixi.toml
moseschmiedel: That doesn't work as we need it for importing the tracing `Zone`s

src/larecs/archetype.mojo
moseschmiedel: I don't see any redundancy here. This is the only place where `set_component` is defined and everything that sets component values uses `set_components` to do so.
The closure is here because of the `VariadicPack.consume_elements()` function

src/larecs/scheduler.mojo
moseschmiedel: Oh, this was an oversight from me

pixi.toml
samufi: yes. The question is whether we could make this import optional. That is, only import if the package is available. If it is not, we could make the zones a no-op. But maybe this is all too complicated.

src/larecs/archetype.mojo
samufi: Sorry, my comment must have been from a previous version. Confirmed. Looks good.

pixi.toml
moseschmiedel: We cannot conditionally import packages in Mojo

samufi (approved):

PR #153, opened 2026-07-17
Disable apply and update mojo

moseschmiedel:
The `Storage` object is just a temporary workaround to not invalidate `world.resources.get[T]()` references when iterating over Queries. As the `World` API will significantly change in the next version (heterogeneous CPU / GPU Systems) this is left here as is to compile correctly but is not yet polished.


scripts/docs.py
```diff
@@ -0,0 +1,269 @@
+import argparse
+import asyncio
+import json
+import platform
+import shutil
+import subprocess
```
samufi: unneded import

scripts/docs.py
```diff
@@ -0,0 +1,269 @@
+import argparse
```
samufi, line 1: Can you quickly tell me why this extra Python module is needed?

benchmark/world/access_benchmark.mojo
```diff
@@ -143,43 +143,43 @@ def benchmark_apply_expexp_1_comp_100_000(
     bencher.iter(bench_fn)
 
 
-def benchmark_apply_simd_expexp_1_comp_100_000(
```
samufi, line 146: Add acomment that this is disabled because of bug XYZ

src/larecs/archetype.mojo
```diff
@@ -86,6 +89,7 @@ struct EntityAccessor[
         with Zone(function_name="EntityAccessor.get_entity()"):
             return self._archetype[].get_entity(self._index_in_archetype)
 
+    @__unsafe_nested_origins_read_only
```
samufi, line 92: Out of interest, what does this do?

src/larecs/storage.mojo
```diff
@@ -0,0 +1,1976 @@
...
+
+    This is always used in combination with a [..world.World] instance, e.g. as a member of it.
+
+    Parameters:
+        ComponentTypes: A variadic list with all possible component types for this storage.
+
+    Examples:
+        ```mojo
+        var world = World[Position, Velocity]()
+        world.storage.add_entitiy(Position(10.0, 20.0)
```
samufi: Missing `)`

src/larecs/storage.mojo
```diff
@@ -0,0 +1,1976 @@
...
+
+    comptime Archetypes = List[Self.Archetype]
+    """Type alias for the list of archetypes owned by this storage."""
+
+    var _archetypes: Self.Archetypes
+    """Storage for all archetypes owned by this storage."""
+
+    var _archetype_map: BitMaskGraph[
+        -1
+    ]  # Mapping from component masks to archetypes.
```
samufi: 2 docstrings

src/larecs/storage.mojo
```diff
@@ -0,0 +1,1976 @@
...
+    """Graph mapping component masks to archetype indices."""
+
+    def __init__(out self):
```
samufi, line 130: missing docstring

test/world_test.mojo
```diff
@@ -579,94 +616,100 @@ def test_world_apply() raises:
...
 
-    assert_false(world.is_locked())
+    assert_false(world.storage.is_locked())
 
 
-def test_world_apply_SIMD() raises:
-    world = SmallWorld()
-    pos = Position(0.0, 2.0)
-    vel = Velocity(0.1, 0.2)
+# def test_world_apply_SIMD() raises:
```
samufi, line 647: Add a comment referring to the issue why this was commented out

samufi (changes requested):
Only small changes requested.

One larger thing: I do not really like the name "storage". However, as there may be larger things changing soon, I am okay with keeping it for now.

scripts/docs.py
moseschmiedel: This is so to have a quick and easy way to build and watch the docs live during development

src/larecs/archetype.mojo
moseschmiedel: This is part of the new interior-origins but is currently not documented and is likely to change. As far as I understand it tells the compiler that the origin of `self` is only read in the method and therefore interior-origins of `self` don't need invalidation. This is the way the stdlib currently implements all the getter methods for containers like `List` or `Dict`.

samufi (approved):

PR #154, opened 2026-07-21
Improve `System` implementation

### LINKED ISSUE #129: GPU-acceleration for Systems
~Add overload for `world.apply` that executes the provided function as a [GPU-Kernel](https://mojolang.org/docs/manual/gpu/fundamentals/#kernel-functions) on the GPU.~

### LINKED ISSUE #144: Make EntityAccessor parametric
EntityAccessor is created by queries, which guarantee that certain components exist. This could (and should) should be exploitet when using the get function of an entity accessor. That way, it can be checked at compile time whether all required components exist. We can still allow get for other cases, but should then call them "unsafe_get". I think this is a much cleaner solution than enabling an assert via a comptime variable (which has not worked for me, btw...)

What do you think @moseschmiedel ?

### LINKED ISSUE #158: Request: make _assert_alive(entity) public
There are use cases where users want to know if an entity still exists (e.g. has a species been removed?). I therefore, _assert_alive should be refactored to assert_alive.

### LINKED ISSUE #159: UB when adding / removing non-trivial components
In `_remove_and_add` in `storage.mojo`, `new_archetype.set_components` is used instead of `new_archetype.init_components`. As a result, the compiler tries to deinit the components at the place before moving in the new components. This results in a double free UB for non-trivial components. Also, the components should not be copied but moved in this action.

I believe the same issue occurs when batch adding and removing. Here, a span is used for setting the new components. `Span.fill` assumes that the values are initialized. Here, we should make an explicit distinction between trivial and non-trivial types and use the built-in functions for moving / memcopying the data. Note that then the clear method needs to be made aware that the data have been moved out already.

A small side note: the `unsafe` marker should be a prefix and not a postfix to methods to follow the Mojo standard.

@moseschmiedel

### LINKED ISSUE #164: Batch operations should accept a Filter directly instead of requiring world.filter[...]()
`HostStorage.query` takes a `Filter` as a comptime parameter, but the batch operations take a runtime `BitMaskFilter`, which the user can only obtain through `World.filter[...]()`:

- `HostStorage.remove_entities` (`src/larecs/host_storage.mojo:600`)
- `HostStorage.add` (`:887`)
- `HostStorage.remove` (`:998`)
- `HostStorage.apply` (`:1509`)
- `Replacer.by` (`:1983`, `:2024`)

So the documented idiom for a batch op is:

```mojo
for entity in world.storage.remove[Velocity](
    world.filter[Filter().include[Position, Velocity]()]()
):
    ...
```

Two nested calls, a hop from `world.storage` up to `world` and back down, and a second filter type the user otherwise has no reason to know about. Next to the query form it looks arbitrary:

```mojo
for entity in world.storage.query[Filter().include[Position, Velocity]()]():
    ...
```

### Suggested fix

Add `[filter: Filter]`-parameterized overloads to the six entry points above, so filters are written the same way everywhere:

```mojo
for entity in world.storage.remove[Velocity, filter = Filter().include[Position, Velocity]()]():
    ...
```

Keep the `BitMaskFilter` overloads for code that builds masks dynamically — `BitMaskFilter` stays the runtime representation, it just stops being the only way in.

At minimum, add a `filter[...]()` helper to `HostStorage` itself. Today it lives only on `World` (`src/larecs/world.mojo:116`), so code holding a `HostStorage` — the receiver of every one of these methods — cannot build the argument they require without also holding the `World`.

Found while reviewing the user-facing `Filter` API as of a2da7c4 (branch `improve-systems`).

moseschmiedel:
Key features:
- Explicit `Component` and `Resource` requirements
- Executable on both GPU and CPU 
  - device-agnostic storage layer
  - builtin chunking (useful for SIMD and GPU warps)
- Conflict detection and resolution between systems


src/larecs/filter.mojo
```diff
@@ -1,67 +1,419 @@
...
-# Filters are required to query entities using [Storage.Query].
-#
-# See [BitMask], [MaskFilter] anf [RelationFilter] for basic filters.
-# For type-safe generics queries, see package [github.com/mlange-42/arche/generic].
-# For advanced filtering, see package [github.com/mlange-42/arche/filter].
+@fieldwise_init
+struct Filter[
+    _include: Components = Components[](),
+    _exclude: Components = Components[](),
+    _is_exclusive: Bool = False,
```
samufi, line 18: Is there a need for `_is_exclusive`? Simply setting everything not included to excluded would be sufficient, wouldn't it? Such an exclude mask could be constructed at comptime if required. Consider removing this parameter so as to reduce complexity.

src/larecs/world.mojo
```diff
@@ -1,59 +1,138 @@
...
+
+        Returns:
+            The number of entities in the world.
         """
         with Zone(function_name="World.__len__(out size: Int)"):
             size = 0
             for archetype in self.storage._archetypes:
                 size += len(archetype)
+
+    def filter[
```
samufi, line 117: This could (and I think should) be a static method. `World.filter(...)` seems to be a valid syntax for me, and `world.filter(...)` will still work.

samufi (approved):
Two minor changes requested. Otherwise fine to merge. We will keep moving fast an can adjust syntax later if need be.

src/larecs/filter.mojo
moseschmiedel: No, `Filter` has no Knowledge about all available `ComponentTypes` by design. So when `Filter.exclusive` is called we cannot construct the `_exclude` set.

samufi: Understood. I thought of it as a bitmask. This was wrong.

src/larecs/world.mojo
moseschmiedel: I also added overloads to all functions taking a `BitMaskFilter` so that we almost never need this function.

(The only use case for this is to create a `Filter` at comptime materialize it to make runtime mutations and then use it for some ECS operation.)

PR #155, opened 2026-07-30
Make Larecs🌲 compile times go brrrrr....

moseschmiedel:
We used a quadratic time component existence check in the `ComponentManager` (in `_ContainsComponents`). This changes the `ComponentManager` to create a `Dict` mapping from `ComponentType` name to `ComponentId` which can then be used for constant time lookup and existence check.


samufi (approved):

moseschmiedel:
@greptileai Please review the PR

PR #157, opened 2026-08-06
Do not invalidate interior resource origins on `Resources.get[T]()`

moseschmiedel:


src/larecs/resource.mojo
```diff
@@ -206,7 +209,11 @@ struct Resources(Copyable, Movable, Sized):
         """
         with Zone(function_name="Resources.get[T: ResourceType]()"):
             try:
-                return self._storage[reflect[T].name()].unsafe_get[T]()
+                return Pointer(
+                    to=self._storage[reflect[T].name()].unsafe_get[T]()
+                )._get_ref_with_unsafe_interior_origin[
+                    "unsafe_box", origin_of(self._storage[reflect[T].name()])
```
samufi: when getting evaluating the origin, the actual key of the dict does not matter. I therefore suspect that all returned references will get the same internal origin. This may not be a problem because of the escape hatch, but without that, this would not work. So, I am not sure how much added value this gives.

samufi:
@moseschmiedel could you also have a look at storage.get? I think this also needs an escape hatch.

src/larecs/resource.mojo
moseschmiedel: Ok, I changed this to `UnsafeAnyOrigin` as we currently cannot model the knowledge we have about the lifetime of a Resource to the compiler.

src/larecs/resource.mojo
```diff
@@ -206,7 +204,9 @@ struct Resources(Copyable, Movable, Sized):
         """
         with Zone(function_name="Resources.get[T: ResourceType]()"):
             try:
-                return self._storage[reflect[T].name()].unsafe_get[T]()
+                return Pointer(
```
samufi, line 207: Just a clarification question: this keeps the resource reference alive until it goes out of scope. Would it also be an option to use different interior origins for the resources?

My solution looked like this:

```mojo
    @__unsafe_nested_origins_read_only
    @always_inline
    def get[
        T: ResourceType
    ](ref self) raises -> ref[
        origin_of(self._storage[reflect[T].name()])._get_owned_interior[
            "unsafe_box"
        ]
    ] T:
        """Gets a resource.

        Parameters:
            T: The type of the resource to get.

        Returns:
            A reference to the resource.
        """
        with Zone(function_name="Resources.get[T: ResourceType]()"):
            try:
                return Pointer(
                    to=self._storage[reflect[T].name()].unsafe_get[T]()
                )._get_ref_with_unsafe_interior_origin[
                    "unsafe_box", origin_of(self._storage[reflect[T].name()])
                ]()
            except DictKeyError:
                raise Error(
                    t"The resource `{reflect[T].name()}` does not exist."
                )
```
Where is this wrong? (I want to learn!)

samufi (commented):
I mainly have a question of understanding. Otherwise the PR is good to go. I'll approve as soon as we have talked about the question.

src/larecs/resource.mojo
moseschmiedel: The `UnsafeAnyOrigin` gives us the ability to keep references to resources valid while calling mutating functions (`def fn(mut self)`) like `remove[T]()` on `Resources`.
Your approach, while allowing multiple refs through `get[T]()` invalidates them on any mutating function call. But we know that `remove[T]()` should never touch memory for any resources beside `T`, so references to other resources must not, and in my opinion should not, be invalidated even if that means we need to use the `UnsafeAnyOrigin` that will get deprecated in the future. Hopefully we have an alternative solution for this use case by then.

Here is a little program showing the difference between the `UnsafeAnyOrigin` and the interior origin:
```mojo
from larecs.unsafe_box import UnsafeBox


struct ReflectedResources:
    var _storage: Dict[StaticString, UnsafeBox]

    def __init__(out self):
        self._storage = Dict[StaticString, UnsafeBox]()

    @always_inline
    def add[T: Copyable & Deinitable](mut self, var value: T) raises:
        self._storage[reflect[T].name()] = UnsafeBox(value^)

    @__unsafe_nested_origins_read_only
    @always_inline
    def get[
        T: Copyable & Deinitable
    ](ref self) raises -> ref[
        origin_of(self._storage[reflect[T].name()])._get_owned_interior[
            "unsafe_box"
        ]
    ] T:
        try:
            return Pointer(
                to=self._storage[reflect[T].name()].unsafe_get[T]()
            )._get_ref_with_unsafe_interior_origin[
                "unsafe_box", origin_of(self._storage[reflect[T].name()])
            ]()
        except DictKeyError:
            raise Error(t"The resource `{reflect[T].name()}` does not exist.")

    def remove[T: Copyable & Deinitable](mut self) raises:
        try:
            _ = self._storage.pop(reflect[T].name())
        except DictKeyError:
            raise Error(t"The resource `{reflect[T].name()}` does not exist.")


struct UnsafeAnyResources:
    var _storage: Dict[StaticString, UnsafeBox]

    def __init__(out self):
        self._storage = Dict[StaticString, UnsafeBox]()

    @always_inline
    def add[T: Copyable & Deinitable](mut self, var value: T) raises:
        self._storage[reflect[T].name()] = UnsafeBox(value^)

    @__unsafe_nested_origins_read_only
    @always_inline
    def get[
        T: Copyable & Deinitable
    ](ref self) raises -> ref[UnsafeAnyOrigin[mut=origin_of(self).mut]] T:
        try:
            return Pointer(
                to=self._storage[reflect[T].name()].unsafe_get[T]()
            ).as_unsafe_any_origin()[]
        except DictKeyError:
            raise Error(t"The resource `{reflect[T].name()}` does not exist.")

    def remove[T: Copyable & Deinitable](mut self) raises:
        try:
            _ = self._storage.pop(reflect[T].name())
        except DictKeyError:
            raise Error(t"The resource `{reflect[T].name()}` does not exist.")


def main() raises:
    var unsafe_any = UnsafeAnyResources()

    unsafe_any.add(42)
    unsafe_any.add[Float32](3.14)

    ref unsafe_int = unsafe_any.get[
        Int
    ]()  # Get a reference to the `Int` resource
    ref unsafe_float = unsafe_any.get[
        Float32
    ]()  # Get a reference to the `Float32` resource
    unsafe_any.remove[
        Float32
    ]()  # Remove the `Float32` resource. Note: no origin invalidation happens here due to UnsafeAnyOrigin
    print(
        unsafe_int
    )  # Access `Int` resource. This works even after removing Float32 due to UnsafeAnyOrigin

    # Sidenote: There is a unsafety built into this approach, because `.remove[T]()` will also not invalidate references to `T`.
    print(
        unsafe_float
    )  # Access `Float32` resource. Sadly this is not caught by the compiler.

    var reflected = ReflectedResources()

    reflected.add(42)
    reflected.add[Float32](3.14)

    ref int = reflected.get[Int]()  # Get a reference to the `Int` resource
    reflected.remove[
        Float32
    ]()  # Remove the `Float32` resource. Note: origin invalidation happens here.
    print(
        int
    )  # Access is not allowed here even though we (the programmer) know that the `Int` resource will always be valid even after calling `.remove[T]()` for any `T != Int`.
```

samufi: Thanks for the explanation!

samufi (approved):

